PR #3, opened 2020-01-20
Add beginning README

dhh1128:


README.md
```diff
@@ -0,0 +1,36 @@
+[![DIDComm logo](didcomm-logo.png)](didcomm-logo.svg)
```
llorllale: where did this logo come from? what is the icon on the bottom right supposed to be? it reminds me a bit of a [moka pot](https://en.wikipedia.org/wiki/Moka_pot)

dhh1128: I made it.

The icon has 3 peers: a person, an institution, and an IoT thing.

PR #28, opened 2020-02-06
propose initial content on routing

dhh1128:


OR13:
We might want to include screenshots of images that have changed in the PR, many reviewers won't know how to checkout the branch to see the updates.

dhh1128:
Can you clarify what you mean by this suggestion, Orie? (100% of the images of changed in this PR,  since no images existed before...)

routing.md
```diff
@@ -1,4 +1,275 @@
...
+
+### Mediators and Relays
+
+DIDComm routing uses two important constructs to model all varieties of network and cryptographic routing: _mediators_ and _relays_.
+
+A __mediator__ is a participant in routing that must be accounted for by the sender's cryptography. In other words, it is visible in the cryptographic routing dimension. It has its own keys and will deliver messages only after decrypting an outer envelope to reveal a `forward` request. It must understand DIDComm routing to do this. Many types of mediators may exist, but two important ones should be widely understood, as they commonly manifest in DID Docs:
+
+1. A service that hosts many agents at a single endpoint to provide herd privacy (sometimes called an "agency") is a _mediator_.
```
TelegramSam: I think `hosts` may have unintended associations. I propose `receives messages for` or `receives messages on behalf of` as a replacement.

routing.md
```diff
@@ -1,4 +1,275 @@
...
+2. A cloud-based agent that forwards messages to mobile devices is a _mediator_.
+
+In contrast, a __relay__ is an entity that passes along encrypted messages without understanding or decrypting them. It can be an off-the-shelf component with little or no concept of special DIDComm routing requirements; it's focused on network routing only. Relays can be used to change the transport for a message (e.g., accept an HTTP POST, then turn around and emit an email; accept a Bluetooth transmission, then turn around and emit something in a message queue). Load balancers and mix networks like TOR are important types of relay.
```
TelegramSam: > Relays can be used to change the transport for a message...

This statement is true, but it is _also_ true for Mediators, which makes it feel out of a place in a sentance starting with `In contrast`.

routing.md
```diff
@@ -1,4 +1,275 @@
...
+        // 0.99.
+        "hop_decay": 0.2,
+        // Likelihood (from 0 to 1) that the message size will
+        // be distorted to the next receiver. 
+        "noise_chance": 0.5
+    }
+}
+```
+
+The value of the `to` field is a DID, not a key. This hides details about the internals of a sovereign domain from external parties. The sender will have had to multiplex encrypt for all relevant recipient keys, but doesn't need to know how routing happens to those keys. The mediator and the receiver may have coordinated about how distribution to individual keys takes place (see [RFC 0211: Route Coordination](https://github.com/hyperledger/aries-rfcs/blob/master/features/0211-route-coordination/README.md)), but that is outside the scope of concerns of this protocol.
```
TelegramSam: using a DID as a value for the `to` field does not identify which of the agents to send the message to. How is that specified?

OR13:
> Can you clarify what you mean by this suggestion, Orie? (100% of the images of changed in this PR, since no images existed before...)

You can screen cap and upload images when you open a PR / Issue. we use this to allow not github uses to quickly see how something has changed, here is an example on an issue:

Screen Shot 2020-02-10 at 12 54 48 PM

TelegramSam:
Not contained in this PR is how a route is represented when passing between recipient and sender. We'll need to align the concepts in how a route functions with `forward` messages to how they are represented.

routing.md
OR13: Feels like maybe it should be a DID URI to an agent...`did:example:123;service=agent` ?

dhh1128: I think you are conflating network routing and cryptographic routing. When Alice sends a message to Bob, she generally doesn't think of this as sending a message to Bob's cell phone specifically; she thinks of it as sending a message to Bob. I suggest that the proper cryptographic routing is to multiplex encrypt for all of Bob's agents that have the privilege of seeing plaintext (as clarified in Bob's DID doc). The message arrives at Bob's mediator attached to a `forward` request where the destination is not a specific Bob agent, but rather Bob's DID. Bob's mediator then forwards to all of Bob's agents that Bob--not Alice--has chosen to receive plaintext messages.

routing.md
dhh1128: Good catch. I will update.

routing.md
dhh1128: Okay. I will update.

dhh1128:
> You can screen cap and upload images when you open a PR / Issue. we use this to allow not github uses to quickly see how something has changed, here is an example on an issue:

Yes, I am aware of the mechanics. What I don't understand is the goal. Github already provides a nice diff view. Your comment leads me to think you're worried about the usability of this feature, specifically for non-Github users. Is that accurate? I just don't understand why a PR submitter should be doing what you advocate. I'm not arguing, just genuinely puzzled.

routing.md
TelegramSam: Fixed.

routing.md
TelegramSam: Fixed

routing.md
```diff
@@ -1,4 +1,277 @@
...
+
+### Remember Routes are One-Way (not Duplex)
+
+In all of this discussion, note that we are analyzing *only* a flow from Alice to Bob. How Bob gets a message back to Alice is a completely separate question. Just because Carl, Darla, Emil, Francis, and Agents-R-Us may be involved in how messages flow from Alice to Bob, does not mean they are involved in flow the opposite direction.
+
+Note how this breaks the simple assumptions of pure request-response technologies like HTTP, that assume the channel in (request) is also the channel out (response). [Duplex](https://en.wikipedia.org/wiki/Duplex_%28telecommunications%29) request-response can be modeled with DIDComm, but doing so requires support that may not always be available, plus cooperative behavior governed by the [`~thread`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0008-message-id-and-threading/README.md) decorator.
+
+### Routing Protocol
```
liormargalit: Would “Forwarding Protocol” title fit better? This section does not explain the routing protocol itself, how the sender or the mediator know where to send the message, but it describes the protocol of sending the forward message once the next node has been selected.

routing.md
kdenhartog: Note, that authcrypt is essentially the same thing as ECDH-1PU which has pointed out that there's cocerns around key compromisation attacks. For more details on this see section 5 of the IETF spec for [ECDH-1PU](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-00#section-5).

PR #30, opened 2020-02-14
Propose initial content on Transports

liormargalit:


transports.md
```diff
@@ -1,8 +1,14 @@
...
-* Transport binding
+## Summary
+This spec details how different transports can be used to transmit DIDComm messages.
+## Motivation
+The DIDComm message is constructed as JWM, using JWE for confidentiality and authentication, so it is transport agnostic.
+Each transport has unique features and properties, so we need to define a standard usage that will be aligned with the asynchronous+simplex design decision.
+## Reference
+### HTTP(S)
+The POST method is used to send the DIDComm message.
+The POST message MIME type is application/didcomm-enc-env, see RFC 0044: DIDComm File and MIME Types for more details.
```
kdenhartog: I'd prefer that we not reference Aries RFCs docs for back history. Rather, I'd prefer that we take out the relevant details to conform with the spec and reference to the spec instead.

kdenhartog: I believe we should modify the MIME type to be able to support multiple forms of DIDComm. Ideally, I'd like to be able to use the MIME type to indicate the format so at least using application/didcomm-enc-env+json would we useful.

liormargalit: Done

liormargalit: If there is a plan to support other format types then we should list it here to indicate all the possible MIME type for DIDComm, but I am aware only of json. Can you please elaborate?

dhh1128: I think we should manage this suggestion as an issue, @kdenhartog. I don't think other MIME type thinking is mature enough to reference yet.

PR #31, opened 2020-02-17
Initial update to connections.

TelegramSam:
Not yet ready for merge, but a huge step closer. Signed-off-by: Sam Curren <telegramsam@gmail.com>

### DIFF
--- a/connections.md
+++ b/connections.md
@@ -40,12 +40,13 @@ The _invitee_ will provision a new DID according to the DID method spec. For a P
   "@type": "https://didcomm.org/didexchange/1.0/request",
   "~thread": { "pthid": "<id of invitation>" },
   "label": "Bob",
-  "connection": {
-    "did": "B.did@B:A",
-    "did_doc": {
-        "@context": "https://w3id.org/did/v1"
-      	// DID Doc contents here.
+  "did": "B.did@B:A",
+  "did_doc~attach": {
+    "mime-type": "",
+    "data": {
+      "base64": "<b64 encoded bytes of diddoc>"// DID Doc contents here.
     }
+    "sig": <sig info>
   }
 }
 ```
@@ -59,7 +60,7 @@ The _invitee_ will provision a new DID according to the DID method spec. For a P
 * The `label` attribute provides a suggested label for the DID being exchanged. This allows the user to tell multiple exchange requests apart. This is not a trusted attribute.
 * The `connection` attribute contains the `did` and `did_doc` attributes. This format maintains consistency with the Response message where this attribute is signed.
 * The `did` indicates the DID being exchanged.
-* The `did_doc` contains the DID Doc associated with the DID. If the DID method for the presented DID is not a peer method and the DID Doc is resolvable on a ledger, the `did_doc` attribute is optional.
+* The `did_doc~attach`  attachment is required when `did` contains a peer did. contains the DID Doc associated with the DID. If the DID method for the presented DID is not a peer method and the DID Doc is resolvable on a ledger, the `did_doc` attribute is optional. The attachment must include a signature
 
 ##### Correlating requests to invitations
 
@@ -80,11 +81,13 @@ When a `request` responds to an implicit invitation, its `~thread.pthid` MUST co
   "@type": "https://didcomm.org/didexchange/1.0/request",
   "~thread": { "pthid": "032fbd19-f6fd-48c5-9197-ba9a47040470" },
   "label": "Bob",
-  "connection": {
-    "did": "B.did@B:A",
-    "did_doc": {
-        // DID Doc contents here.
+  "did": "B.did@B:A",
+  "did_doc~attach": {
+    "mime-type": "",
+    "data": {
+      "base64": "<b64 encoded bytes of diddoc>"// DID Doc contents here.
     }
+    "sig": <sig info>
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


TelegramSam:
@dhh1128 This last update includes DIDDocs as attachments (for Peer DIDs only). I believe that is what you needed for Peer DIDs, correct?

Also, I was without an example for how signed attachments would work, so I left that to fill out in the future.

TelegramSam:
Now ready for merge with TODOs in place

connections.md
```diff
@@ -1,8 +1,252 @@
...
+  "@type": "https://didcomm.org/didexchange/1.0/request",
+  "~thread": { "pthid": "<id of invitation>" },
+  "label": "Bob",
+  "did": "B.did@B:A",
+  "did_doc~attach": {
+    "mime-type": "",
+    "data": {
+      "base64": "<b64 encoded bytes of diddoc>"// DID Doc contents here.
+    }
+    "sig": <sig info>
```
OR13: signature of what? if this is a signature of the encoded did document... this data structure is very similar to initial-state ion sidetree.

connections.md
```diff
@@ -1,8 +1,252 @@
...
+
+##### Example
+
+```json
+{
+  "@id": "5678876542345",
+  "@type": "https://didcomm.org/didexchange/1.0/request",
+  "~thread": { "pthid": "<id of invitation>" },
+  "label": "Bob",
+  "did": "B.did@B:A",
```
OR13: I'd prefer to see `did:example:123`, or `did:example:bob` and `example.com` or peer-id, etc...

connections.md
```diff
@@ -1,8 +1,252 @@
...
+An invitation is presented in one of two forms:
+
+* An explicit OutOfBand invitation with its own `@id`.
+* An [implicit](#implicit-invitation) invitation contained in a DID document's [`service`](https://w3c-ccg.github.io/did-spec/#service-endpoints) attribute.
+
+When a `request` responds to an explicit invitation, its `~thread.pthid` MUST be equal to the `@id` property of the invitation.
+
+When a `request` responds to an implicit invitation, its `~thread.pthid` MUST contain a [DID URL](https://w3c-ccg.github.io/did-spec/#dfn-did-url) that resolves to the specific `service` on a DID document that contains the invitation.
+
+TODO: Include examples using a ledger resolvable DID.
```
OR13: wondering if we will be needing lots of real world examples, if we should choose to use `did:web` or something like that so its easy for us to update...

connections.md
```diff
@@ -1,8 +1,252 @@
...
+```
+
+
+##### Request Transmission
+
+The Request message is encoded according to the standards of the Encryption Envelope, using the `recipientKeys` present in the invitation.
+
+If the `routingKeys` attribute was present and non-empty in the invitation, each key must be used to wrap the message in a forward request, then encoded in an Encryption Envelope. This processing is in order of the keys in the list, with the last key in the list being the one for which the `serviceEndpoint` possesses the private key.
+
+The message is then transmitted to the `serviceEndpoint`.
```
OR13: Good to add an example here, something like `https://cloud-agent.example.com/...`

connections.md
```diff
@@ -1,8 +1,252 @@
...
+
+The signature data must be used to verify against the invitation's `recipientKeys` for continuity.
+
+##### Attributes
+
+* The `@type` attribute is a required string value that denotes that the received message is an exchange request.
+* The `~thread` block contains a `thid` reference to the `@id` of the request message.
+* The `connection` attribute contains the `did` and `did_doc` attributes to enable simpler signing.
+* The `did` attribute is a required string value and denotes DID in use by the _inviter_. Note that this may not be the same DID used in the invitation.
+* The `did_doc~attach` attribute contains the associated DID Doc. If the DID method for the presented DID is not a peer method and the DID Doc is resolvable on a ledger, the `did_doc` attribute is optional.
```
OR13: is the use of `~attach` significant here? I remember seeing something about this syntax in previous aries RFCs... if this is already in use, I'm very much in favor of not changing it and creating extra work, just want to make sure I know when to suggest the use of `~`.

dhh1128: The ~attach stuff is a decorator naming convention. We'll probably be revising decorators in Aries into message headers in DIDComm, so this convention may become less relevant before we're done.

connections.md
TelegramSam: sig details are included as a TODO.

TelegramSam:
@OR13 Are you ok with these items being handled post-merge? Do you think we need each recognized in an explicit TODO before we proceed?

OR13:
@TelegramSam Absolutely, sorry this was hanging open.

connections.md
OR13: Ok makes sense I'm gonna resolve this comment.

connections.md
```diff
@@ -1,8 +1,252 @@
...
+##### Correlating requests to invitations
+
+An invitation is presented in one of two forms:
+
+* An explicit OutOfBand invitation with its own `@id`.
+* An [implicit](#implicit-invitation) invitation contained in a DID document's [`service`](https://w3c-ccg.github.io/did-spec/#service-endpoints) attribute.
+
+When a `request` responds to an explicit invitation, its `~thread.pthid` MUST be equal to the `@id` property of the invitation.
+
+When a `request` responds to an implicit invitation, its `~thread.pthid` MUST contain a [DID URL](https://w3c-ccg.github.io/did-spec/#dfn-did-url) that resolves to the specific `service` on a DID document that contains the invitation.
```
liormargalit: Have the usage of timestamps and/or nonces been considered (to prevent replay attack) in case of implicit invitation?

dhh1128: @liormargalit : That is an important issue. The good news is that every message in DIDComm includes a unique @id property. This is essentially a nonce. Thus, there is no need to include an additional nonce field to prevent replay.

liormargalit: @dhh1128  thank you for the reply, but this is correct for explicit invitation, while for implicit invitation, the spec defines that "~thread.pthid MUST contain a DID URL that resolves to the specific service on a DID document that contains the invitation". There is no "nonce" value to correlate with the invitation.

kdenhartog: Of note, if we want to treat `@id` as an nonce in some cases (I think it's fine) then we may need to strengthen the language around the implementation to prevent implementation mistakes. What I'm thinking about here is while a UUIDv4 satisfies the requirement for id today, a person may choose to set `@id` sequentially such that it's predictable to generate a replay attack. 

To improve it saying something like `@id` must be a UUIDv4 would suffice.

PR #34, opened 2020-03-01
feat: jwm profile definition

tplooker:
The following PR is a draft that starts the discussion on how best to integrate JWM into DID comm messaging.

Based on the comments from @troyronda and @awoie on the last call in regards to supported JWAs we will need to discuss the `MUSTs` in the specification further and their impact on interoperability.

The most obvious outstanding questions for me are around
- The expression of the supported cryptographic algorithms by a DID subject, such as in their DID document or in an invitation to connect. E.g I must be-able to tell the other party what cryptographic algorithms I support so they can prepare a message for me.
- Are we supporting key wrapping and if so, are we supporting all the content encryption ciphers with key wrapping?

### DIFF
--- a/jwm.md
+++ b/jwm.md
@@ -1,0 +1,97 @@
+# DIDComm Messaging
+
+DIDComm messaging, is a secure messaging protocol built atop of [decentralized identifiers](https://w3c.github.io/did-core/). DIDComm messages are JSON based leveraging the secure messaging format of JWM (JSON Web Message) which is a [IETF draft](https://github.com/mattrglobal/jwm) to expand the [JOSE](https://datatracker.ietf.org/group/jose/documents/) (Javascript object signing and encryption) family of specifications for the purposes of secure messaging.
+
+## Terminology
+
+**Message Payload** In reference to JWMs, `message payload` refers to the [JWM Attribute Set](https://tools.ietf.org/html/draft-looker-jwm-01#section-1.2).
+
+## JWM Profile
+
+The following section defines a JWM profile for DIDComm messages. This profile defines the usage of the JWM attributes defined by the [specification](https://tools.ietf.org/html/draft-looker-jwm-01#section-3.1) as well as defines some attributes that are unique to DIDComm messaging.
+
+- **id** - REQUIRED. Message ID. The `id` attribute value MUST be assigned in a manner that ensures that there is a negligible probability that the same value will be accidentally assigned to another DIDComm message.
+- **type** - REQUIRED. Message Type. The `type` attribute value MUST be a valid [MTURI](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/uris.md#mturi), that when resolved gives human readable information about the message. The attributes value also informs the content of the message, for example the presence of other attributes and how they should be processed.
+- **to** - OPTIONAL. Recipient(s) identifier. The `to` attribute MUST be an array of strings where each element is a valid [DID](https://w3c.github.io/did-core/#generic-did-syntax) which identifies the recipients of the message.
+- **from** - OPTIONAL. Sender identifier. The `from` attribute MUST be a string that is a valid [DID](https://w3c.github.io/did-core/#generic-did-syntax) which identifies the sender of the message.
+- **reply_url** - OPTIONAL. Reply url. The `reply_url` attribute value MUST be a string that is a valid [URL](https://tools.ietf.org/html/rfc3986) which identifies where the sender wishes the recipient(s) to send a reply.
+- **reply_to** - OPTIONAL. Reply to. The `reply_to` attribute value MUST be an array of strings where each element is a valid [DID URL](https://w3c.github.io/did-core/#generic-did-syntax) which identifies the recipients of the reply to the message.
+- **created_time** - OPTIONAL. Message Created Time. The `created_time` attribute is used for the sender to express when they created the message.
+- **expires_time** - OPTIONAL. Message Expired Time. The `expires_time` attribute is used for the sender to express when they consider the message to be expired.
+
+### Example JWM Message Payload
+
+```
+{
+    "id":"ef5a7369-f0b9-4143-a49d-2b9c7ee51117",
+    "type":"https://didcomm.org/didexchange/1.0/invitation",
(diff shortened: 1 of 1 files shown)


tplooker:
@awoie a clarification I'd like to make is that the crypto listed here is not meant to be an exhaustive list, merely the minimum baseline to support interoperability. We need to decide what is required vs what is optional to support.

awoie:
> @awoie a clarification I'd like to make is that the crypto listed here is not meant to be an exhaustive list, merely the minimum baseline to support interoperability. We need to decide what is required vs what is optional to support.

@tplooker I understand that this is not an exhaustive list. If we cover secp256k1 as well, then we have a pretty good representation of the crypto that our community uses. Ed25519/X25519 for signing/key agreement and secp256k1/secp256r1 for signing were the algorithms most people used. I'm just questioning whether it is reasonable to suggest secp256k1/r1 for key agreement as well as there might be some security implications with these elliptic curve domain parameters. However, if both are considered to be similar in terms of security, then I would add support for both. I'm not a cryptographer but reading that sheet https://safecurves.cr.yp.to/twist.html it looks quite similar.

awoie:
@christianlundkvist do you have some thoughts on that?

mikelodder7:
> The expression of the supported cryptographic algorithms by a DID subject, such as in their DID document or in an invitation to connect. E.g I must be-able to tell the other party what cryptographic algorithms I support so they can prepare a message for me.

This has been the plague of many security systems when you negotiate ciphers. That's usually 90% of the security holes right there. If possible, I would negotiate, I would say only support 1 cipher and version it. If you move to a new version then change the cipher. This might be harsh, but that's been many of the bugs in OpenSSL and its cousins, Kerberos, LDAP, and many others.

> Are we supporting key wrapping and if so, are we supporting all the content encryption ciphers with key wrapping?

I would say pick one that you want to support and just do one. Edge clients must be able to speak all of the ones they want with which they want to be interoperable, but recipients should just do one.

mikelodder7:
> I'm just questioning whether it is reasonable to suggest secp256k1/r1 for key agreement as well as there might be some security implications with these elliptic curve domain parameters. 

K-256, P-224..P-521 with Key agreements are virtually the same. I would suggest using x963 as the standard here. As the cryptographer **Thomas Pornin** says "Formally, ANSI X9.63 defines use of elliptic curve Diffie-Hellman, and the 'shared secret' is obtained by hashing the X coordinate of the resulting point. The Bitcoin world, intent on doing things its own way with no regard for standards (after all, at its core, Bitcoin is an anarchist endeavour), decided to do things differently. Since Bitcoin is the main user of secp256k1, their way of doing things is a de facto standard."

Curve25519 is really nice because all Elliptic Curve operations are constant time––implementers don't have to do anything special code-wise to make it constant time. This applies to x25519 and Ed25519. You do not have this guarantee with K-256, P-256..P-521. If you don't think this is a problem see [CopyCat](https://arxiv.org/pdf/2002.08437.pdf) and [TpmFail](https://tpm.fail/tpmfail.pdf).
The other downside to ECDSA is its dependence on the RNG when signing, which means yet another factor to be aware of. See [Biased Nonce Sense](https://eprint.iacr.org/2019/023.pdf) if you want to worry about your BitCoin and Ether being safe. Ed25519 doesn't need a TRNG during the signing process, only during key generation.

K256 has the advantage of being a Koblitz curve which means the operations are in a binary field vs a prime field yielding performance improvements over say the NIST P curves.

The NIST P curves have the advantage of just that, they are approved by NIST and have been for quite some time. [FIPS 186-5](https://csrc.nist.gov/publications/detail/fips/186/5/draft) plans to add K256 and Curve25519 which effectively voids this advantage when it becomes official.

awoie:
@tplooker @OR13 Currently, we have a discussion on which crypto should be supported in didcomm-messaging. I am challenging the requirement that all DID methods will have to support both ECDH-ES (P-256) + AES256GCM AND X25519 + XChacha2020Poly1305? 
- what if my HSM does not support both?
- what if i don't want to do the key management for both types of keys?
- the size of the DID Doc needs also to be considered?
- In case of Blockchain-based DIDs, you would also need bigger transactions or in our case, two separate transactions to add the keys. 
- limitations for IOT devices also need to be considered. I saw client devices with 256kb of memory and having them to support additional crypto might not be feasible.

I'm ok with having both supported in general (as an option), but i won't mandate support for both. 

I also want to emphasise that I am very unhappy that ECDSA with curve K-256 for signing is not explicitly mentioned. That is a clear requirement from the more Blockchain-centric DID community (e.g., ETH, Bitcoin).

tplooker:
@awoie, I understand and wont resist us taking the approach of recommending the different crypto that should be supported but falling short of defining what MUST be supported. To that list we can discussing adding support for secp256k1 however the more we add, the harder interoperability gets.

OR13:
IMO, there should be 1 mandatory and it should probably be:  X25519 + XChacha20Poly1305

The rest should be optional.

The more optionality, the worse the spec, the harder the interop.

I understand the desire to be able to use did comm with limited ledger transactions for methods such as did:ethr or did:btcr.... but I don't think we should weaken the entire spec because of their built in support for secp256k1 alone... 

I think it's 1 blockchain operation needed to add support for ed25519 and x25519, if we are smart about how those keys are related.

The fact that an operation is required could be viewed as a good thing... we would all have a lot less spam, if emails cost money to send.

tplooker:
@OR13 interested to consider what the nature of the normative statements will be for secure data storage as it is essentially the same trade offs at play here and the same crypto?

kdenhartog:
Realistically we can mandate this, but it will lead to more headaches with companies that have FIPS requirements. Either they’ll choose not to support it compliantly or they won’t support it at all.

OR13:
Its worth noting that node12 uses openssl and ships with support for both FIPS and non FIPS crypto... so adding MUST to a none FIPS, and MAY to a FIPS is fine... endless discussion could be had on this subject... https://safecurves.cr.yp.to/

IMO, best thing to do in recommend ed25519, and allow p-384.

You can tell you are right when everyone hates your decision :)

kdenhartog:
Is from optional now that we're using JWKs in the outer layer?

PR #37, opened 2020-03-26
Correct DID Exchange version.

TelegramSam:
Existing Aries work (source material) listed the version as 1.0. Bumping a major version is necessary to avoid conflicts with the changes we are making.

### DIFF
--- a/connections.md
+++ b/connections.md
@@ -39,7 +39,7 @@ Example
 ```json
 {
   "@id": "5678876542345",
-  "@type": "https://didcomm.org/didexchange/1.0/request",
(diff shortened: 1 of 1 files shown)


TelegramSam:
Discussed in WG Call 2020-03-30. No comments were offered.

kdenhartog:
I'm for your proposal to make everything in this version of didcomm start at 2.0

kdenhartog:
Rereading my comment now that this has been merged. I was conflating VC exchange with DID exchange again, so the answer to that question is... Doesn't matter.

PR #39, opened 2020-03-27
Updated the Exchange Protocol with challenge step

liormargalit:
Added challenge step in case of request following an implicit invitation to prevent replay attack.
This issue was noted in the comments of https://github.com/decentralized-identity/didcomm-messaging/pull/31

### DIFF
--- a/connections.md
+++ b/connections.md
@@ -190,7 +190,61 @@ Possible reasons:
 
 **response_processing_error**
 
-- unknown processing error## 3. Exchange Acknowledgement
+- unknown processing error
+
+#### 3. Exchange Challenge
+
+The exchange challenge message is used as the response for request following an implicit invitation. This message is required to merge the two flows with the sequence of an explicit invitation. In particular, the explicit invitation consists of unique `@id` to prevent replay attack.
+
+##### Example
+
+```json
+{
+  "@type": "https://didcomm.org/didexchange/1.0/challenge",
+  "@id": "12345678901234567",
+  "~thread": {
+    "thid": "<The Thread ID is the Message ID (@id) of the first message in the thread>"
+  },
+  "did": "<did inviter is sending to invitee>"
+}
+```
+
+##### Attributes
+
+* The `@type` attribute is a required string value that denotes that the received message is a challenge request.
+* The `~thread` block contains a `thid` reference to the `@id` of the request message.
+* The `did` attribute is a required string value and denotes DID in use by the _inviter_. Note that this may not be the same DID used in the invitation.
+
+In addition to a new DID, the associated DID Doc might contain a new endpoint. This new DID and endpoint are to be used going forward in the relationship.
+
+##### Response Transmission
+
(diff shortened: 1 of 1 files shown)


liormargalit:
@TelegramSam  @dhh1128 
For some reason I cannot set reviewers in the pull request.
Can you please review?

TelegramSam:
I don't understand the problem this is solving. The text once mentions replay attack prevention, but nothing else. Can you elaborate a little on the problem and how this solves it?

liormargalit:
@TelegramSam  
This solves the issue raised in the comment in https://github.com/decentralized-identity/didcomm-messaging/pull/31 
In order to protect the inviter from replay attack, the exchange request that responds to an explicit invitation include the id of the invitation in its ~thread.pthid.
Currently there is no protection in case of an implicit invitation, so the challenge step is added.

TelegramSam:
RIGHT. Sorry for forgetting that bit of context. Given that, what is the expected flow? 
Implicit Invitation -> Request -> Challange -> ?

liormargalit:
@TelegramSam 
The expected flow is 
Implicit invitation -> Request -> Challenge -> Request with the same properties as in case of explicit invitation (the id of the challenge in its ~thread.pthid).

dhh1128:
I'm still confused, although I'm grateful for @liormargalit 's most recent comments, because they are beginning to give me confidence.

It seems to me that an implicit invitation is inherently replayable. I'm imagining, for example, a QR code on the front of a soccer stadium, inviting the world to connect with the team. Alice can accept that invitation and connect, and she can do so 1 time or 200 times, because the invitation is intended to be replayable. So what replay attack are we trying to mitigate?

swcurran:
Further - I thought an implicit invitation was where the inviter has a public DID and the invitee sends a `request` using the DIDDoc info for that public DID.  Is that correct?  A publisher of such a DID should expect such requests, and so I'm not sure of the replay attack in that scenario.

The example I'm thinking of is a holder has proven claims from a credential to a verifier, and the verifier does not know/recognize the issuer.  They might connect to inquire under what authority are they issuing the credential.

TelegramSam:
> It seems to me that an implicit invitation is inherently replayable 

This is where my head is. I want to understand the nature of the problem before I can evaluate the solution.

liormargalit:
The scenario is that the attacker will manage to replay Alice's request and the inviter will drop the connection with Alice in order to establish the new connection. The inviter cannot differentiate between the request sent by the attacker and the case that Alice sends the same request multiple times.

TelegramSam:
Alice (invitee) provides an unique message ID inside of each request. A replay would have the same message ID. If the invitee does innocently send multiple requests each would have it's own id.

Is that not sufficient coverage for a request replay?

liormargalit:
Closing this pull request. Will add a new section to the spec about reliable messaging with suggestion to use the timestamp in the JWM message payload to prevent replay attack.

TelegramSam:
Thank you @liormargalit for your work on this topic.

PR #40, opened 2020-03-30
content update for scope

TelegramSam:
Pulled primarily from source documents and adapted.


TelegramSam:
Discussed in WG Call 2020-03-30. No comments were offered.

PR #44, opened 2020-05-08
Message Structure and Protocol URIs

TelegramSam:
This work is a partial application of the DIDComm profile of jwm (also in the repo) and some included details about protocol type URIs that related.

### DIFF
--- a/message_structure.md
+++ b/message_structure.md
@@ -9,7 +9,7 @@ The following example shows common elements of a DIDComm Message. Further detail
     "@id": "1234567890",
     "@type": "<message-type-uri>",
     "from": "did:example:alice",
-    "to": "did:example:bob",
+    "to": ["did:example:bob"],
     "created_time": "1516269022",
     "expires_time": "1516269022",
     "body": {
--- a/message_structure.md
+++ b/message_structure.md
@@ -1,0 +1,31 @@
+## Message Structure
+
+DIDComm Messages are based on JWM (JSON Web Messages). A message has a basic structure that specifies the message type, id, and other attributes common to all messages. A message also includes attributes specific to the message type. Prior to being sent to a recipient, the JWM is encrypted into a JWE according to the JWM spec.
+
+The following example shows common elements of a DIDComm Message. Further details and advanced usage are covered elsewhere in this spec.
+
+```json
+{
+    "@id": "1234567890",
+    "@type": "<protocolURI>/<protocolversion>/<messagetype>",
+    "from": "did:example:alice",
+    "to": "did:example:bob",
+    "created_time": "1516269022",
+    "expires_time": "1516269022",
+    "body": {
+    	"messagespecificattribute": "and it's value"
+	}
+}
+```
+
+
+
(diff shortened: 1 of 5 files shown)


message_structure.md
```diff
@@ -0,0 +1,31 @@
...
+```
+
+
+
+The attributes of a DIDComm Message are as follows:
+
+- **id** - REQUIRED. Message ID. The `id` attribute value MUST be unique to the sender.
+- **type** - REQUIRED. Message Type. The `type` attribute value MUST be a valid [MTURI](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/uris.md#mturi), that when resolved gives human readable information about the message. The attributes value also informs the content of the message, for example the presence of other attributes and how they should be processed.
+- **to** - OPTIONAL. Recipient(s) identifier. The `to` attribute MUST be an array of strings where each element is a valid [DID](https://w3c.github.io/did-core/#generic-did-syntax) which identifies the recipients of the message.
+- **from** - OPTIONAL. Sender identifier. The `from` attribute MUST be a string that is a valid [DID](https://w3c.github.io/did-core/#generic-did-syntax) which identifies the sender of the message. For DID methods that use query parameters to carry additional information, they might also be present in the from string.
```
kdenhartog, line 28: I think we have to require this now to get did linkage because it's not coming in the outer envelope anymore.

TelegramSam: I was going to address this in the section about encryption, but I agree that a mention here is also useful.

TelegramSam:
Discussed in WG Call 2020-05-11. Updated to attribute to be a list instead of a singular string.

awoie:
@TelegramSam this PR seems uncontroversial, I think we can merge this.

message_structure.md
```diff
@@ -0,0 +1,31 @@
...
+	}
+}
+```
+
+
+
+The attributes of a DIDComm Message are as follows:
+
+- **id** - REQUIRED. Message ID. The `id` attribute value MUST be unique to the sender.
+- **type** - REQUIRED. Message Type. The `type` attribute value MUST be a valid [Message Type URI](protocols.md#message-type-uri) , that when resolved gives human readable information about the message. The attributes value also informs the content of the message, for example the presence of other attributes and how they should be processed.
```
peacekeeper, line 26:
```suggestion
- **@id** - REQUIRED. Message ID. The `@id` attribute value MUST be unique to the sender.
- **@type** - REQUIRED. Message Type. The `@type` attribute value MUST be a valid [Message Type URI](protocols.md#message-type-uri) , that when resolved gives human readable information about the message. The attributes value also informs the content of the message, for example the presence of other attributes and how they should be processed.
```

TelegramSam: Good catch! I believe this should be updated the other way - by removing the `@` prefix in the examples. This does have implications about JSON-LD compatibility that we have not discussed. I'll put this specific topic on the agenda for Monday's meeting and we can update appropriately to the decisions there.

PR #48, opened 2020-05-09
Added Service Endpoint Details

TelegramSam:

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -288,7 +288,7 @@ DIDComm DID Document endpoints have the following format:
 {
     "id": "did:example:123456789abcdefghi#didcomm-1",
     "type": "DIDComm",
-    "publicKey": "did:example:123456789abcdefghi#key-1",
+    "publicKeys": ["did:example:123456789abcdefghi#key-1"],
     "serviceEndpoint": "http://example.com/path",
     "routingKeys": ["did:example:somemediator#somekey"]
 }
@@ -298,7 +298,7 @@ DIDComm DID Document endpoints have the following format:
 
 **type**: MUST be `DIDComm`. 
 
-**publicKey**: MUST specify the key to be used when encrypting a message to be sent to this endpoint.
+**publicKey**s: MUST be an array of strings, to specify the keys to be used when encrypting a message to be sent to this endpoint.
 
 **serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec.
 
--- a/routing.md
+++ b/routing.md
@@ -67,11 +67,11 @@ In contrast, a __relay__ is an entity that passes along encrypted messages witho
 Like mediators, relays can be used to change the transport for a message (e.g., accept an HTTP POST, then turn around and emit an email; accept a Bluetooth transmission, then turn around and emit something in a message queue). But unlike mediators, relays can do this without understanding DIDComm. Load balancers and mix networks like TOR are important types of relay.
 
 Let's define mediators and relays by exploring how they manifest in a series of communication scenarios between Alice and Bob.
- 
+
 #### Scenario 1: direct
 
 Alice and Bob are both employees of a large corporation. They work in the same office, but have never met. The office has a rule that all messages between employees must be encrypted. They use paper messages and physical delivery as the transport. Alice writes a note, encrypts it so only Bob can read it, puts it in an envelope addressed to Bob, and drops the envelope on a desk that she has been told belongs to Bob. This desk is in fact Bob's, and he later picks up the message, decrypts it, and reads it.
- 
+
 ![routing scenario 1: direct](/collateral/routing-scenario-1.png)
 
 In this scenario, there is no mediator, and no relay.
@@ -79,7 +79,7 @@ In this scenario, there is no mediator, and no relay.
 #### Scenario 2: a gatekeeper
 
 Imagine that Bob hires an executive assistant, Carl, to filter his mail. Bob won't open any mail unless Carl looks at it and decides that it's worthy of Bob's attention.
- 
+
 Alice has to change her behavior. She continues to package a message for Bob, but now she must account for Carl as well. She take the envelope for Bob, and places it inside a new envelope addressed to Carl. Inside the outer envelope, and next to the envelope destined for Bob, Alice writes Carl an encrypted note: "This inner envelope is for Bob. Please forward."
 
 ![routing scenario 2: mediator](/collateral/routing-scenario-2.png)
@@ -103,9 +103,9 @@ Like scenario 3, Darla brings Bob his mail at home. However, Bob isn't at home w
 ![routing scenario 4: many relays](/collateral/routing-scenario-4.png)
 
 Here, Emil and Francis are also acting as relays. Note that *nobody knows about the full route*. Alice thinks she's delivering directly to Bob. So does Darla. Bob knows about Darla and Emil, but not about Francis.
- 
+
 Note, too, how the transport is changing from physical mail to email to text.
- 
+
 To the party immediately upstream (closer to the sender), a relay is indistinguishable from the next party downstream (closer to the recipient). A party anywhere in the chain can insert one or more relays upstream from themselves, as long as those relays are not upstream of another named party (sender or mediator).
 
 ##### More Scenarios
@@ -184,7 +184,7 @@ There are 3 roles in the protocol: `sender`, `mediator`, and `recipient`. The se
 >Note: When the mediator is the routing agent of a single identity subject like Alice, the logical receiver is Alice herself, but the physical receiver may manifest as multiple edge devices (a phone, a laptop, a tablet). From the perspective of this protocol, multiplexing the send from mediator to receiver is out of scope for interoperability--compatible and fully supported, but not required or specified in any way.
 
 In this protocol, the sender and the receiver never interact directly; they only interact via the mediator.
-  
+
 The sender can decorate the `forward` message in standard DIDComm ways: using [`~timing.expires_time`, `~timing.delay_milli` and `~timing.wait_until_time`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0032-message-timing/README.md#tutorial) to introduce timeouts and delays, and so forth. However, the mediator is NOT required to support or implement any of these mixin semantics; only the core forwarding behavior is indispensable. If a mediator sees a decorator that requests behavior it doesn't support, it MAY return a [`problem-report`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0035-report-problem/README.md) to the sender identifying the unsupported feature, but it is not required to do so, any more than other recipients of DIDComm messages would be required to complain about unsupported decorators in messages they receive.
 
 >One particular decorator is worth special mention here: [`~please_ack`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0015-acks/README.md#requesting-an-ack-please_ack). This decorator is intended to be processed by ultimate recipients, not mediators. It imposes a burden of backward-facing communication that mediators should not have. Furthermore, it may be used to probe a delivery chain in a way that risks privacy for the receiver. Therefore, senders SHOULD NOT use this, and mediators SHOULD NOT honor it if present. If a sender wishes to troubleshoot, the [message tracing](https://github.com/hyperledger/aries-rfcs/blob/master/features/0034-message-tracing/README.md) mechanism is recommended.
@@ -268,7 +268,7 @@ The `mix` property is discussed next.
 Normally, the payload attached to the `forward` message received by the mediator is transmitted directly to the receiver with no further packaging. However, optionally, the mediator can attach the opaque payload to a new `forward` message, which then acts as a fresh outer envelope for the second half of the delivery. This [rewrapping](#rewrapping) means that the "onion" of packed messages stays the same size rather than getting smaller as a result of the forward operation:
 
 ![re-wrapped sequence](/collateral/routing-roles-2.png)
- 
+
 Rewrapping mode is invisible to senders, but mediators need to know about it, since they change their behavior as a result. Receivers also need to know about it, because it causes them to receive a double-packaged message instead of a singly-packaged one. The outer envelope is a `forward` message where `to` is the receiver itself.
 
 Why is such indirection useful?
@@ -277,9 +277,35 @@ Why is such indirection useful?
 * It lets the mediator remain uncommitted to whether the next receiver is another mediator or not. This may provide flexibility in some routing scenarios.
 * It lets the mediator change the size of the message by adding or subtracting noise from the content. 
 * It allows for dynamic routing late in the delivery chain.
- 
+
 These last two characteristics are the foundation of mix networking feature for DIDComm. That feature is the subject of a different RFC; here we only note the existence of the optional feature. 
 
 
-### DID Docs and Routing
+### DID Document Service Endpoint
+DIDComm DID Document endpoints have the following format:
+
+```json
+{
+    "id": "did:example:123456789abcdefghi#didcomm-1",
+    "type": "DIDComm",
+    "publicKey": "did:example:123456789abcdefghi#key-1",
+    "serviceEndpoint": "http://example.com/path",
+    "routingKeys": ["did:example:somemediator#somekey"]
+}
+```
+
+**id**: must be unique, as required in [DID Core](https://www.w3.org/TR/did-core/#service-endpoints). No special meaning should be inferred from the `id` chosen.
+
+**type**: MUST be `DIDComm`. 
+
+**publicKey**: MUST specify the key to be used when encrypting a message to be sent to this endpoint.
+
+**serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec.
+
+**routingKeys**: An optional list of keys to be used when preparing the message for transmission as specified in the [Routing] section of this spec.
+
+#### Multiple Endpoints
+
+A DID Document may contain multiple service entries of type `DIDComm`. Unless otherwise specified, messages should be sent to endpoints in the order they are specified.
+
 [TODO: discuss how routing info is exposed in a DID doc, and how it is conveyed in "ephemeral mode" where no DID doc is available.]
--- a/routing.md
+++ b/routing.md
@@ -282,7 +282,9 @@ These last two characteristics are the foundation of mix networking feature for 
 
 ### DID Document Keys
 
-TODO: include details about how DIDComm keys are represented/identified in the DID Document.
+All keys declared in the DID Document's `keyAgreement` section should be used to encrypt messages. The details of key representation are described in the [Public Keys section of the DID Core Spec](https://www.w3.org/TR/did-core/#public-keys).
+
+TODO: include details about how DIDComm keys are represented/identified in the DID Document. The DID Core Spec appears light on details and examples of `keyAgreement` keys. Clarifying language should be included here or there.
 
 ### DID Document Service Endpoint
 
--- a/routing.md
+++ b/routing.md
@@ -280,15 +280,18 @@ Why is such indirection useful?
 
 These last two characteristics are the foundation of mix networking feature for DIDComm. That feature is the subject of a different RFC; here we only note the existence of the optional feature. 
 
+### DID Document Keys
+
+TODO: include details about how DIDComm keys are represented/identified in the DID Document.
 
 ### DID Document Service Endpoint
+
 DIDComm DID Document endpoints have the following format:
 
 ```json
 {
     "id": "did:example:123456789abcdefghi#didcomm-1",
     "type": "DIDComm",
-    "publicKeys": ["did:example:123456789abcdefghi#key-1"],
     "serviceEndpoint": "http://example.com/path",
     "routingKeys": ["did:example:somemediator#somekey"]
 }
@@ -298,11 +301,9 @@ DIDComm DID Document endpoints have the following format:
 
 **type**: MUST be `DIDComm`. 
 
-**publicKey**s: MUST be an array of strings, to specify the keys to be used when encrypting a message to be sent to this endpoint.
-
 **serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec.
 
-**routingKeys**: An optional list of keys to be used when preparing the message for transmission as specified in the [Routing] section of this spec.
+**routingKeys**: An optional ordered array of strings referencing keys to be used when preparing the message for transmission as specified in the [Routing] section of this spec. 
 
 #### Multiple Endpoints
 
--- a/routing.md
+++ b/routing.md
@@ -284,6 +284,8 @@ These last two characteristics are the foundation of mix networking feature for 
 
 All keys declared in the DID Document's `keyAgreement` section should be used to encrypt messages. The details of key representation are described in the [Public Keys section of the DID Core Spec](https://www.w3.org/TR/did-core/#public-keys).
 
+Keys used in a signed JWM are declared in the DID Document's `authenication` section.
+
 TODO: include details about how DIDComm keys are represented/identified in the DID Document. The DID Core Spec appears light on details and examples of `keyAgreement` keys. Clarifying language should be included here or there.
 
 ### DID Document Service Endpoint


TelegramSam:
I'm not sure why it complains. The text view (within Github) renders this perfectly.

routing.md
```diff
@@ -277,9 +277,35 @@ Why is such indirection useful?
...
 
-### DID Docs and Routing
+### DID Document Service Endpoint
+DIDComm DID Document endpoints have the following format:
+
+```json
+{
+    "id": "did:example:123456789abcdefghi#didcomm-1",
+    "type": "DIDComm",
+    "publicKey": "did:example:123456789abcdefghi#key-1",
```
troyronda, line 291: Although I understand this has been the DIDComm convention from Aries, I am curious about specifying publicKey in this way.

- I've seen others using "verification method" sections in the DID document to specify key usage. e.g., `keyAgreement` seems similar to this usage.
- Is this defined the same way as public keys defined in "verification method" sections? Example: https://www.w3.org/TR/did-core/#authentication

routing.md
```diff
@@ -277,9 +277,35 @@ Why is such indirection useful?
...
- 
+
 These last two characteristics are the foundation of mix networking feature for DIDComm. That feature is the subject of a different RFC; here we only note the existence of the optional feature. 
 
 
-### DID Docs and Routing
+### DID Document Service Endpoint
+DIDComm DID Document endpoints have the following format:
+
+```json
```
troyronda, line 287: To have maximum interoperability, I suspect that we should consider a JSON-LD context hosted from DIF that contains custom vocabulary.

TelegramSam:
I updated the `publicKey` attribute to be a list of keys (`publicKeys`). The discussion about this attribute existing at all is still open.

routing.md
TelegramSam: Agree. After we decide on the needed attributes, a PR to include such information would be useful.

routing.md
TelegramSam: @dhh1128 is this consistent with your argument about not designating keys in the service block?

routing.md
```diff
@@ -277,9 +277,35 @@ Why is such indirection useful?
...
+{
+    "id": "did:example:123456789abcdefghi#didcomm-1",
+    "type": "DIDComm",
+    "publicKeys": ["did:example:123456789abcdefghi#key-1"],
```
OR13, line 291: Is there a more descriptive term that we could use here? for what purpose are these keys linked to a service endpoint?

routing.md
```diff
@@ -277,9 +277,35 @@ Why is such indirection useful?
...
+    "serviceEndpoint": "http://example.com/path",
+    "routingKeys": ["did:example:somemediator#somekey"]
```
OR13, line 293: IMO, this is specified correctly, as long as a context defines `routingKey` somewhere.

routing.md
troyronda: Sounds good - seems like a followup topic.

routing.md
```diff
@@ -277,9 +277,35 @@ Why is such indirection useful?
...
+
+**id**: must be unique, as required in [DID Core](https://www.w3.org/TR/did-core/#service-endpoints). No special meaning should be inferred from the `id` chosen.
+
+**type**: MUST be `DIDComm`. 
+
+**publicKey**s: MUST be an array of strings, to specify the keys to be used when encrypting a message to be sent to this endpoint.
+
+**serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec.
+
+**routingKeys**: An optional list of keys to be used when preparing the message for transmission as specified in the [Routing] section of this spec.
```
troyronda, line 305: Is this still an ordered list?

dhh1128:
I vote that we either omit entirely, or strongly discourage, the use of `publicKey` (or whatever it is renamed to) within this block.

The default behavior in DIDComm should be to send to the controller of a __DID__, NOT to send to a specific key that the controller of the DID is using in a particular place. The controller of a DID should be able to change the configuration of their internal infrastructure, including which keys are active behind a particular endpoint, without having to update their DID doc. It is not something the sender should worry about. The sender, instead, should always encrypt for every recipient key that's eligible to receive the message (where "eligible" means "declared or referenced in a verification method section"), no matter what endpoint they're using.

To @troyronda 's point above, we should declare something analogous to a verification method section at the root of the DID doc to list keys that are eligible to receive encrypted communication. I say "analogous to a verification method" because what goes into this section is the opposite of a verification method; it's __inbound receipt__ of messages instead of __outbound actions__ like other verification methods. But it looks like a verification method anyway:

```jsonc
{
    // ...other sections of DID doc omitted...
    // a section for the "capabilityInvocation" verification method
    "capabilityInvocation" [ ... ]
    // a section for the "keyAgreement" verification method; keys that will be used to encrypt outbound comms from DID controller go here
    "keyAgreement" [ ... ]
    // the new section I'm recommending
    "readEncrypted" [
        // either references to keys defined elsewhere, or inlined definitions of keys, go here.
    ]
    "service": [
        { // a service block structured as Sam proposed, minus the "publicKey" part that I claim is a mistake. 
        } // Sender always encrypts for every key that appears in readEncrypted, no matter what service
          // endpoint is used.
    ]
}
```

routing.md
troyronda: There is an example in the DID core spec for Identity Hub that also has the publicKey reference.

It seems the higher-level question is: am I defining the DIDComm publicKey for a particular agent or am I defining "my" DIDComm publicKeys in general.

There seem to be examples of both in the DID core spec.
For the latter, see above comment. For the former, see the identity hub example in https://www.w3.org/TR/did-core/#example-16-various-service-endpoints

So both usages likely have their place.

dhh1128:
A simplification of what I recommended just above would be to conflate `keyAgreement` and `readEncrypted`, such that we always encrypt for everything that appears in `keyAgreement`. I didn't advocate that because I believe it's possible for keys to be legitimate senders but not legitimate receivers -- but perhaps I could be talked out of that intuition.

routing.md
dhh1128: See my comment [below](https://github.com/decentralized-identity/didcomm-messaging/pull/48#issuecomment-626964010).

dhh1128:
> I'm not sure why it complains. The text view (within Github) renders this perfectly.

It looks to me like the code blocks are JSON but have interleaved comments. Using "jsonc" instead of "json" as the code block type should allow us to interleave `// comment` stuff.

routing.md
dhh1128: It MUST be. Otherwise the correct order for wrapping cannot be determined. And this requires a custom JSON-LD context, since JSON-LD insists on viewing arrays as unordered by default (contrary to the JSON spec).

troyronda: (this can have considerations for JSON-LD DID documents.)

kdenhartog:
Rather than putting this at the top level, could we instead add a section to the service endpoint that states the keys that should be used for the service endpoint? In that case, the keys would be a subset of the keys used in "keyAgreement" which doesn't require us to define an extension.

dhh1128:
@kdenhartog : that's the very behavior I claim is an antipattern. There shouldn't be a strong association between an endpoint and a subset of keys. Period.

kdenhartog:
> 
> 
> @kdenhartog : that's the very behavior I claim is an antipattern. There shouldn't be a strong association between an endpoint and a subset of keys. Period.

Can you explain why this is an antipattern? To me it seems like a pattern to enforce privilege of least authority.

dhh1128:
I left some comments about this in DIF slack. Reproduced in part here:

1. There is (should be) information asymmetry between Alice and the outside world, as far as her disposition of keys is concerned. Alice needs to know that her laptop and her tablet are both capable of receiving messages via the agency service she subscribes to, while her iPhone is not. The rest of the world shouldn’t care. Therefore, making Alice declare this info in a public doc makes no sense. It’s like asking Alice to define all her NATing rules inside her DNS CNAME record. It’s bad for privacy, causes unnecessary churn of the DID doc, and makes senders be fancy. What senders should do, 99% of the time, is simply encrypt for all keys that Alice has said are legitimate encryption targets. Alice can then fiddle with whether she wants to add her iPhone to her agency, changing mind right in the middle of conversations if she likes, without any concerns about fragility or senders having to track her updated configs.

2. The set of keys usable via an endpoint isn’t actually an endpoint property, anyway. That’s just a convenient simplification, approximately true at a snapshot in time. Suppose Alice has an http:// endpoint and an smtp:// endpoint. Do we really think she’s going to create separate keys for each — or is she going to use one endpoint to transform (relay) into another transport? And what about the always-possible relay that could be used with any transport, which is saving the message to a log or DB somewhere for later debugging or auditing? Should senders plan for that? Should Alice have to tell them to do so? If they didn’t plan for that, are we willing to live with the idea of asking the sender to re-send with different encryption? No, I think what we really want is to say that when you send to Alice, you send to her DID, and encrypt for all the keys that are authorized to decrypt for her, no matter which endpoint you use. The only exception to this would be if Alice asks you to send to a specific key — and she should do that in a message/protocol, not in her DID doc.

OR13:
I think for performance, we need some way of communicating that a given service endpoint does not support some of the keys in the did document... or we need to assert that all keys in a did document are always available to agents registered in service endpoints, and some how links agent DIDs to other user DIDs.

When creating an agent, its a best practice to generate keys in a hardware isolated environment, and then register them to the DID... 

Reusing keys is generally considered to be an anti pattern (although like reusing passwords.... its exceedingly common without software assistance).

This means that if there are multiple agents, they will have disjoint sets of keys.

either the mapping must be communicated in the did document, or it can be communicated as a set of linked DIDs... for example, I might generate a DID per agent, and then link them to myself... similar to how an Executive Assistant and Lawyer might both receive sensitive information on behalf of their customer.... in those cases we have knowledge of a social graph, and we encrypt for a target DID, knowing that they can process requests on behalf of clients... this would seem to be in line with Daniels comments.

If we want to only have a single DID, with multiple agents, I don't see any choice but to handle the mapping of keys to service endpoints.... or take an approach of using all keys to encrypt every message (performance hit).

dhh1128:
>for performance, we need some way of communicating that a given service endpoint does not support some of the keys in the did document

There is negligible performance cost to encrypt for multiple keys, because the only thing you encrypt uniquely for each recipient is the symmetric content encryption key. If the main message is, say, 20 KB in size, and symmetric keys are 128 bits or 16 bytes, then you are encrypting 20 KB once, and 16 bytes as many times as you have different recipients.

>Reusing keys is generally considered to be an anti pattern (although like reusing passwords.... its exceedingly common without software assistance).

I emphatically agree. My recommendation has NOTHING to do with key reuse.

>either the mapping must be communicated in the did document, or it can be communicated as a set of linked DIDs...

No. The mapping doesn't have to be communicated at all. It has to be *known* -- to the DID controller. It doesn't need to be known to the outside world.

dhh1128:
Nobody thinks the mental model for sending email messages is, "I am sending an email message to Alice's iPhone, OR I am sending it to Alice's laptop, OR I am sending it to Alice's tablet." The mental model is, "I am sending it to Alice, and she can decide without me where she wants to read it."

Why do we think it's a value-add to complicate that for both the sender and the receiver, in DIDComm?

swcurran:
@dhh1128 - my understanding is that the recipient tells the sender "send messages here". The sender has no idea what "here" is - they just do what the recipient tells them.  They encrypt the message to the private key(s) of the recipient that are "here".  Then, they also wrap the messages in wrappers to all of the mediators along the way - the mediators that (again) the recipient told them about.

That's the way DIDComm was designed and I'd like to ensure that is still the model we're after, or if that is changing. If a change is being proposed, what is the change?

From what I can tell, the only thing being debated is whether the "here" is defined explicitly in the service entry suing key references or keys, or inferred from other parts of the DIDDoc.

dhh1128:
>They encrypt the message to the private key(s) of the recipient that are "here". 

This sentence is the nub of the difference. I am saying that there are no private keys "here." That is not a construct we need or want. There are just private keys that can decrypt, having no location. The mental model should be that a sender encrypts with as many keys as the DID controller tells them to, no matter where messages are sent. Those keys are declared explicitly in the DID doc; there's no need to infer them. They should just not be declared under an endpoint.

I am happy to expose service endpoints as a location ("here"). And I am happy to send to a DID. What I am not happy about is the idea that endpoints have keys, from the perspective of a sender. Endpoints may have keys, but that's a detail for the receiver to know, and nobody else. That's desirable for privacy reasons, and for DID doc churn reasons, and for simplicity of sender code reasons, and for architectural encapsulation reasons.

I'm not aware of any examples where a key needs to be associated with a DIDComm endpoint to accomplish a use case. Can you think of any?

Orie's pushback about performance was concrete. I don't accept it as a general concern, but I think it could become interesting in rare circumstances (e.g., an agent for an institution that has hundreds of keys). However, in such a case, I suspect the institution would often be partitioned into independent agents (e.g., one agent for Acme's HR dept, another for its IT dept, etc). The way to handle this is not to declare keys for specific endpoints; it's to use different DIDs, or to reply to a first message in an interaction in such a way that you say, "for the rest of this interaction, please reply to key X only." That's not a service endpoint property -- it's a property of a specific interaction. It's the equivalent of calling general tech support for Acme, and being given Fred in the tech support department as the handler of your case. That's quite different from Acme listing and maintaining all its tech support agents when it publishes its general tech support number. When you call in, you don't care who you're talking to, and Acme reserves the right to late-bind who can pick up the phone.

dhh1128:
Here are some scenarios that are either awkward or impossible with the current thinking about binding the keys at a service endpoint early and statically:

* Alice receives messages through endpoint X and routes them to her iPhone. Bob sends Alice a message. Meanwhile, Alice's iPhone dies, but she picks up her tablet and fetches the message and reads it there, instead. (Note that this scenario is more or less equivalent to "Alice changes her DID doc after Bob caches it; Bob encrypts only for a key Alice is no longer using at that endpoint.")

* Alice wants to rotate the keys that she uses at an endpoint without allowing her key usage patterns to fingerprint her.

* Alice wants to rotate the keys that she uses at an endpoint without allowing herself to be correlated temporally.

* Alice wants to guarantee that all of her messages are readable by an auditor, no matter what DIDComm service endpoints come and go in her DID doc. No messages are ever routed to the auditor; rather, the auditor key is on paper stored in a vault somewhere. (This is doable by duplicating the auditor key in every service point definition, but it's an annoyance and unnecessarily verbose.)

OR13:
So in summary, @dhh1128 your proposal is always encrypt for all keys listed, and send messages to the endpoint, and let the receiver handle decryption if they can, on the device they are using?

I would register a hardware backed cloud key for my cloud agent, a hardware backed mobile key for my mobile agent, some auditor keys for my auditor...

A sender would encrypt for all these, and I would be able to read on any device, because of the redundant encryption.

When I'm compromised, and a new key is added, messages will automatically get encrypted for it... we might want to add some UX on the sender side to mitigate this... similar to how signal handles key rotation... but thats a separate issue (on the sender side).

A side effect of always encrypting for all possible keys is that you now have to be very sure those keys are safe to use, and you will likely audit them better.... which I like.

dhh1128:
>So in summary, @dhh1128 your proposal is always encrypt for all keys listed, and send messages to the endpoint, and let the receiver handle decryption if they can, on the device they are using?

Yes.

>A sender would encrypt for all these, and I would be able to read on any device, because of the redundant encryption.

There's an important nuance missing. The recipient would be able to read on any device __that has one of the listed keys__. Not on any device without the constraint. And no keys are ever copied anywhere.

>When I'm compromised

You do the same thing under my regime that you do under the alternate proposal: update your DID doc to remove the compromised key. The difference is that under my proposal, you only have to do it in one place, and someone watching your DID doc can't tell as much about you as a result of the rotation.

An imaginary benefit of the keys-under-endpoint approach is that key rotation is more granular because you can rotate for a single compromised endpoint. I say "imaginary" because this evaporates under scrutiny. If an endpoint is compromised, only the routing key for the endpoint (mediator) has to be rotated. If an agent key is compromised, it has to be rotated in every endpoint through which it's usable. And that turns out to be the same behavior under either proposal.

swcurran:
I'm looking at how this was done in Aries, which is defined [here](https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0094-cross-domain-messaging#bobs-did-for-his-relationship-with-alice), including an example DIDDoc.

The proposal is simply to drop the `recipientKeys` item from the service endpoint and encrypt the recipient message for all of the keys in the `publicKey` array above. Correct?

If that is the case, how are the mediator keys handled, where I need to reference them in the `routingKeys` array, but I (probably) don't want them included in the (eliminated) `recipientKeys` array?

dhh1128:
>The proposal is simply to drop the recipientKeys item from the service endpoint and encrypt the recipient message for all of the keys in the publicKey array above. Correct?

No. The proposal is to drop `recipientKeys` from the service endpoint and encrypt the recipient message for all of the keys in "some array" which might be `keyAgreement`, might be something else, but would most definitely NOT be the generic `publicKey` array.

Routing keys are not declared as keys that are targets for encrypting plaintext; they are only targets for encrypting forward messages. They remain in a `routingKeys` array and are used exactly as described by the current verbiage in the DIDComm spec, this PR, and Aries.

swcurran:
Thanks.  How do we determine where "some array" is?  Wait for the DID spec. to settle?

I scanned this document and I don't see where the `routingKeys` are defined, only referenced.  Do you anticipate them being defined in the DIDDoc and referenced in the array (as in the example in Aries RFC), or being `did:key`s or being DIDs (also an example in the Aries RFC), or some/all of the above?

dhh1128:
>How do we determine where "some array" is? Wait for the DID spec. to settle?

We describe the "verification method" we want. If it matches `keyAgreement`, we use that. If it doesn't, we christen a new one with a name we choose, and make sure it shows up as a canonical extension to DID docs, with the appropriate JSON-LD context, etc. [This issue](https://github.com/w3c/did-core/issues/276) against DID core probably needs to get resolved.

I think `routingKeys` has to be very flexible, deliberately. So I wouldn't be surprised if all of the mechanisms you listed end up being valid. I have suggested that a more appropriate name for these items is "mediators", since they could be DIDs as well as keys, and they are not the only things involved in routing. But that's a whole 'nuther conversation.

kdenhartog:
So are you saying that we'll have two buckets? One bucket contains an ordered list of keys to be able to route a message and this bucket is in the did document. The second is a bucket of keys that can decrypt the ciphertext and view the actual message.

If we do it this way (not opposed, just figuring out your view better at this point and poking some edge cases) what assumptions are we making about keys that are intended for other services that could cause all applications to encrypt to the same keys whether they're using DIDComm or some other application service?

dhh1128:
>So are you saying that we'll have two buckets? One bucket contains an ordered list of keys to be able to route a message and this bucket is in the did document. The second is a bucket of keys that can decrypt the ciphertext and view the actual message which is also in the did document.

Close, but let me nit-pick a bit.

I'm suggesting that we should have a single bucket at the DID doc level says, "Here are keys you should encrypt for when you talk to this DID by DIDComm." Whenever you talk to Alice's DID over DIDComm, you encrypt for all the keys listed in this bucket.

Underneath each service endpoint, there is one instance of a second bucket type which contains externally visible mediators for that endpoint. (This is the JSON array currently named `routingKeys`, but what it really contains is an ordered list of externally visible mediators. Calling these things "routing keys" is inaccurate, since they could be keys, DID key references, or DIDs of some other entity that mediates for the current DID. It is also inaccurate because there could be plenty of other routing things going on; this is just the externally visible mediators, which are the only hops the sender needs to know about to achieve delivery). For each item in this list, the sender wraps a message in a forward.

>what assumptions are we making about keys that are intended for other services that could cause all applications to encrypt to the same keys whether they're using DIDComm or some other application service?

If they share the idea of reading encrypted communications, then they should share use of the array. Please note, however, that "read encrypted data" is the opposite of a verification method, because it doesn't prove anything to someone on the outside; rather, someone on the outside is proving something to the DID controller. See this [issue on DID core](https://github.com/w3c/did-core/issues/276), which is somewhat complementary to the one you logged, and where I am waiting for comment.

I agree that proof purposes should not be application-specific. However, the "don't put it at the top level" advice contradicts other advice that Dave gave in did-core #273, [here](https://github.com/w3c/did-core/issues/273#issuecomment-624292040).

OR13:
So is `keyAgreement` sufficient for DIDComm uses?

I'm convinced by @dhh1128 that these mappings need not be advertised externally.

dhh1128:
I think `keyAgreement` *might* be sufficient for the JWEs used by DIDComm. I don't think it would be the right verification method for the JWSes. Does that seem correct?

Re. my use of the word "might": I would prefer not to invent a new thing, but I have two hesitancies.

1. `keyAgreement` in other contexts is framed as an outbound thing. The DID controller is essentially saying: "This is the method I will use to arrive at a shared symmetric key, so I can prove I'm who I claim to be when I speak outbound." The wrinkle that DIDComm adds is that at serviceEndpoints, we're actually talking about at least an opposite phenomenon, if not an altogether different one: "These are the keys you should encrypt for, if you want to talk inbound to me." That's not a proof purpose or a verification method in the sense given by the current DID spec. I pointed out this odd dissonance in [this issue on DID core](https://github.com/w3c/did-core/issues/276). I think I could get past this hesitation, especially if we broadened the DID spec's definition of a verification method a little. 

2. I'm not sure that it's always true that the same set of keys is used to speak outbound and receive inbound. It's quite possible to conceive of an agent that only speaks, or an agent that only listens. Giving a speak-only agent the ability to read plaintext, or giving a listen-only agent the ability to talk back, feels like a cybersecurity mistake. I'm not sure what to do with this concern. Like I said, I'd prefer not to use something new -- but I don't want to become less secure to do so, and I also don't want to wrest the intended meaning of what's currently there.

dhh1128:
To expand a bit on my second concern: I believe the speak-only or listen-only scenario may be more common in institutional and IoT cases. For example, suppose NASA has various agents. Some are primitive ones hooked to sensor feeds; all they do is publish to the world the latest data from a particular space mission vehicle. The keys for such agents would have to appear in `keyAgreement`, since they are officially endorsed emitters of encrypted data from NASA. Now suppose NASA also has conversations going on with its suppliers/vendors, and it wants to conduct a negotiation in secret about prices. Whatever agents NASA uses for these conversations clearly need inbound keys, able to receive encrypted communication from the outside world. If those keys are defined to be the same list as `keyAgreement`, then I worry we are creating an oddness that hackers could exploit; a hacker would gain access to encrypted inbound communication if it can steal a sensor key. And I think NASA is likely to protect sensor keys at a lower level of cybersecurity from the ones they use to interact with vendors.

But my feeling on this is more intuition than strongly held reasoning. Please poke at my assumptions.

TelegramSam:
This conversation is exactly the kind of conversation I was hoping we'd have as a result of discussing these details. In my last PR update I removed the `publicKeys` attribute from the service block, and added a section with a TODO to include details about how keys are represented elsewhere in the DID Document.

Please comment on the PR as now represented, and I invite PRs for the newly created section.

@dhh1128 @kdenhartog @OR13 @swcurran

OR13:
When sharing a document using encrypted data vaults I believe `keyAgreement` is used as an inbound thing... for example, I lookup your did document, I encrypt my data for you, and I add an access capability for you... you invoke that capability, download the document, and decrypt the data.

I think it would be a huge win to have DIDComm use `keyAgreement` in much the same way... I think it will work for JWEs.

Regarding JWSs... the proper verificationMethod is likely `authentication`. 

If these 2 were chosen, they would likely be supported out of the box by most DID Methods today.

dhh1128:
I'm very happy with using `authentication` for JWSes.

Regarding `keyAgreement`: I see the value of using something familiar. But please help me get past my concern about cybersecurity.

routing.md
```diff
@@ -277,9 +277,36 @@ Why is such indirection useful?
...
+}
+```
+
+**id**: must be unique, as required in [DID Core](https://www.w3.org/TR/did-core/#service-endpoints). No special meaning should be inferred from the `id` chosen.
+
+**type**: MUST be `DIDComm`. 
+
+**serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec.
+
+**routingKeys**: An optional ordered array of strings referencing keys to be used when preparing the message for transmission as specified in the [Routing] section of this spec.
```
OR13, line 306: I think we can use the array syntax, and define a JSON-LD context that says that `routingKeys` are an `@list`.... and the desired behavior will be achieved for both representations:

https://json-ld.org/spec/latest/json-ld/#example-51-an-ordered-collection-of-values-in-json-ld

OR13:
keyAgreement is currently defined here:

- https://w3c-ccg.github.io/security-vocab/#keyAgreement
- https://w3c.github.io/did-core-registries/#keyagreement

In both cases, I interpret it as a bucket for "Public key encryption / ECDH"...

https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.3

Added support for X25519...

My reading of https://en.wikipedia.org/wiki/Curve25519

Combined with the fact that https://w3c.github.io/did-core-registries/#x25519keyagreementkey2019 is currently the only defined `keyAgreement` type...

leads me to believe that keyAgreement either currently means, or will in the future (with sufficient contribution to vocabulary) mean DH capable keys.

https://tools.ietf.org/html/rfc8037

I would expect this collection (`keyAgreement`) to contain such keys... and currently, I would not expect such keys anywhere else in a did document.

@dhh1128 can you restate your question about cybersecurity?

I'm interpreting it as "Alice and Bob might want 2 sets of ECDH capable keys"... I believe that is more of an application level concern, its possible to encrypt for both keys... that fact that Alice might plan to only receive on 1 and send from another is up to her... she is technically capable of doing that if she has 2 keys, but she can also just use 1 to talk with Bob if she wants... AFAIK, the security of the system is still just trust in the ECDH implementation, having 2 keys doesn't strengthen the system... it might leak meta data about which direction a message is going... but Im not enough an expert on ECDH.... paging @tplooker

dhh1128:
My cybersecurity concern is that we may be conflating two activities that are best kept separate: the emitting of secure communications, and the receiving of secure communication. Both require DH to negotiate a session key, so everything you're saying about `keyAgreement` makes sense to me. But as a practical matter and a human one, I'm not sure all keys that are capable of secure communication should be assumed to be bi-directional (used to listen AND used to speak).

In "real life", it's common to have a mismatch between who can listen authoritatively, and who can speak authoritatively. For example, a job applicant named Bob engages with a company about an opening. The hiring manager, Alice, is his point of contact. She interviews him; she is the official recipient when he sends info to the company about his references, etc. But when she decides to hire him, she calls him and says, "I've made the decision, but you'll have to wait for something official until I can get the VP, Carol, to sign off on your hire letter." We do this kind of thing because speaking with authority on behalf of an org is higher risk than listening with authority.

A proof purpose or verification method of `keyAgreement` collapses this distinction. If we use it, we would be saying, "if you have a key that's capable of DH, then you can use it to establish a secure communications channel for this DID, and you can both listen and talk over that channel in all cases. It's one privilege to do both things; we can't distinguish between them." Effectively, Alice the hiring manager can always sign Bob's offer letter; there's nothing in the model that lets the company require a signature from the VP, Carol.

When `keyAgreement` is viewed purely as a proof purpose or verification method as currently defined, this issue is out of scope, because neither proof purposes nor verification methods concern themselves with passive abilities that manifest when triggered by external actors; they concern themselves with the DID controller doing something active toward the outside world. You're invoking something, signing something, asserting something, delegating something... that the world has to confirm or accept. Listening doesn't seem to fit that paradigm very well.

OR13:
Exposing a `keyAgreement` or any ECDH capable key is expressing a cryptographic interface... the closest technology that is in use today is OpenPGP and Key Servers.... AFAIK people tend not to make the distinction you are making in that community... 

We can't prevent someone from encrypting for a key that supports it... if you don't want to receive encrypted messages, don't register the key... if you want to send encrypted messages, and you register a key, you are automatically registering a key which can be used to receive... 

I think the specifics around protocols, or patterns of use that require multiple keys  / split listening and speaking... are a different layer, and there is nothing wrong with enforcing behavior at that layer... I don't think there is a concept of an ECDH key that is expressed specifically for listening or receiving outside of an "implied application context" like I made the IoT platform work that way... and in any case, the device is always capable of doing both if it expresses an ECDH capable key... the question is should we invent a new bucket which communicates that although this is an ECDH key, I will only listen, and never speak from it.

In my opinion, we should not do that, and just use `keyAgreement` and conventions at the application layer.

dhh1128:
Okay. Your example of ssh keys makes me comfortable that `keyAgreement` is an acceptable answer. We'll handle the concern I raised at a different layer.

So keys that create JWSes in DIDComm have to be inlined or referenced in the `authentication` section, and keys that use DH to derive a symmetric key for encrypting (ie.g., emit or receive JWEs from/for them) are in the `keyAgreement` session. Any other opinions, or should we close this thread out?

dlongley:
I agree with @OR13's [comment](https://github.com/decentralized-identity/didcomm-messaging/pull/48#issuecomment-628204677) on this.

TelegramSam:
Related question: if keyAgreement keys are to be used for DIDComm Encryption, AND other purposes (such as SDS), Are we concerned about identifying the use cases? Is the issue of encrypting to extra keys (not used for DIDComm) a concern here?

(I'm low on opinions on the right thing to do, but I want to understand it clearly.)

swcurran:
As I've been thinking about this, I think we can ignore the multi-purpose keyAgreement concern. Agent implementations that use did:peer for DIDComm, will have single purpose DIDs, so no issue.  For receivers that use multi-purpose DIDs with multiple keys in keyAgreement, the receiver's message distributor component must know the rules for distributing each message instance. If only some are of the keyAgreement keys are for DIDComm, the non-DIDComm messages can be ignored. There will be some unneeded work for the sender, but presumably not much.

TelegramSam:
I've added just a bit of language to represent what appears to be the conclusion here. Any objections to merging it?

dhh1128:
I am happy with the language that got added. I won't vote against a merge, but I would still prefer to have `routingKeys` renamed.

routing.md
TelegramSam: resolving, as this has been removed.

routing.md
TelegramSam: resolving, as `publicKeys` has been removed.

routing.md
TelegramSam: resolving, as `routingKeys` has been removed.

TelegramSam:
@dhh1128 I don't consider the naming between `routingKeys` and `mediators` to be resolved, but that discussion I think will be better informed with an upcoming PR. I propose we merge this one as is, having discussed and resolved a major important aspect.

dhh1128:
Okay, that's fine. Let's get this merged.

PR #49, opened 2020-05-09
Add Queue as a Transport option

TelegramSam:
Discussion that led to this addition: https://github.com/hyperledger/aries-rfcs/issues/405

### DIFF
--- a/transports.md
+++ b/transports.md
@@ -9,6 +9,7 @@ Each transport has unique features and properties, so we need to define a standa
 The POST method is used to send the DIDComm message.
 The POST message MIME type is application/didcomm-enc-env.
 A successful response code is 202 (Accepted), but 200 (OK) is also allowed.
+
 #### WebSocket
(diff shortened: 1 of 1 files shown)


transports.md
```diff
@@ -17,3 +18,11 @@ TODO: Include details of the websocket message to define a standard usage.
 TODO:
 * Quic
 * Transport binding
+
+#### Queue
```
kdenhartog, line 22: This seems a bit odd to describe it as a transport. It seems like a pattern in how transports are used rather than an actual transport method itself. I agree that this needs to be described within this spec because it's a key function of routing to mobile devices, but let me see if there's a better place for this to live to prevent confusion about what the "transport layer" actually is.

TelegramSam: Agree it doesn't it one aspect of transports (because it isn't an actual transport.) This section is also used to understand how to process the various endpoint types in a DIDComm service entry of a DID Document. In that aspect, it fits right in.

This is on the list of conversation topics for today's call.

TelegramSam:
I'm going to pull close this PR at this point and revisit it later, after routing and some of the other issues have been more thoroughly discussed.

PR #52, opened 2020-05-18
Adds an encryption section

TelegramSam:
Based off material in the DIDComm JWM Profile Doc. Signed-off-by: Sam Curren <telegramsam@gmail.com>

### DIFF
--- a/encryption.md
+++ b/encryption.md
@@ -11,17 +11,19 @@ When a sender would like to encrypt a message for a recipient or multiple recipi
 
 For the keys involved in key agreement, the following elliptic curves MUST be supported.
 
-|Curve |Description                                                                                                                        |
-|------|-----------------------------------------------------------------------------------------------------------------------------------|
-|X25519|The underlying curve is actually `Curve25519`, however when used in the context of diffie helman the identifier of `X25519` is used|
-|P-256 |NIST defined P-256 elliptic curve                                                                                                  |
+| Curve  | Description                                                  |
+| ------ | ------------------------------------------------------------ |
+| X25519 | The underlying curve is actually `Curve25519`, however when used in the context of Diffie-Hellman the identifier of `X25519` is used |
+| P-256  | NIST defined P-256 elliptic curve                            |
 
 For content encryption of the message, the following algorithms MUST be supported.
 
-|Algorithm(JWA) |Description                                          |
-|---------------|-----------------------------------------------------|
-|C20P           |ChaCha20Poly1305                                     |
-|AES-GCM        |Advanced Encryption Standard with galois/counter mode|
+| Algorithm(JWA) | Description                                           |
+| -------------- | ----------------------------------------------------- |
+| XC20P          | XChaCha20Poly1305                                     |
+| AES-GCM        | Advanced Encryption Standard with galois/counter mode |
+
+TODO: Include language about safe nonce considerations.
 
 ### Anonymous Encryption
 
@@ -29,17 +31,19 @@ When a sender would like to encrypt a message for a recipient or multiple recipi
 
 For the keys involved in key agreement, the following elliptic curves MUST be supported.
 
-|Curve |Description                                                                                                                        |
-|------|-----------------------------------------------------------------------------------------------------------------------------------|
-|X25519|The underlying curve is actually `Curve25519`, however when used in the context of diffie helman the identifier of `X25519` is used|
-|P-256 |NIST defined P-256 elliptic curve                                                                                                  |
+| Curve  | Description                                                  |
+| ------ | ------------------------------------------------------------ |
+| X25519 | The underlying curve is actually `Curve25519`, however when used in the context of Diffie-Hellman the identifier of `X25519` is used |
+| P-256  | NIST defined P-256 elliptic curve                            |
 
 For content encryption of the message, the following algorithms MUST be supported.
 
-|Algorithm(JWA) |Description                                          |
-|---------------|-----------------------------------------------------|
-|C20P           |ChaCha20Poly1305                                     |
-|AES-GCM        |Advanced Encryption Standard with galois/counter mode|
+| Algorithm(JWA) | Description                                           |
+| -------------- | ----------------------------------------------------- |
+| XC20P          | XChaCha20Poly1305                                     |
+| AES-GCM        | Advanced Encryption Standard with galois/counter mode |
+
+TODO: Include language about safe nonce considerations.
 
 ### Examples
 
--- a/encryption.md
+++ b/encryption.md
@@ -18,10 +18,10 @@ For the keys involved in key agreement, the following elliptic curves MUST be su
 
 For content encryption of the message, the following algorithms MUST be supported.
 
-| Algorithm(JWA) | Description                                           |
-| -------------- | ----------------------------------------------------- |
-| XC20P          | XChaCha20Poly1305                                     |
-| AES-GCM        | Advanced Encryption Standard with galois/counter mode |
+| Algorithm(JWA) | Description                |
+| -------------- | -------------------------- |
+| XC20P          | XChaCha20Poly1305          |
+| A256GCM        | AES-GCM with a 256 bit key |
 
 TODO: Include language about safe nonce considerations.
 
@@ -38,10 +38,10 @@ For the keys involved in key agreement, the following elliptic curves MUST be su
 
 For content encryption of the message, the following algorithms MUST be supported.
 
-| Algorithm(JWA) | Description                                           |
-| -------------- | ----------------------------------------------------- |
-| XC20P          | XChaCha20Poly1305                                     |
-| AES-GCM        | Advanced Encryption Standard with galois/counter mode |
+| Algorithm(JWA) | Description                |
+| -------------- | -------------------------- |
+| XC20P          | XChaCha20Poly1305          |
+| A256GCM        | AES-GCM with a 256 bit key |
 
 TODO: Include language about safe nonce considerations.
 
--- a/encryption.md
+++ b/encryption.md
@@ -1,0 +1,48 @@
+## Message Encryption
+
+DIDComm supports two types of message encryption: Authenticated Sender Encryption and Anonymous Sender Encryption. Both forms are encrypted to the recipient, but only Authenticated Sender Encryption provides assurances of who the sender is. 
+
+The encrypted form of a JWM is a JWE. The JOSE family defines [JSON Web Algorithms](https://tools.ietf.org/html/rfc7518) (JWAs) which standardize certain cryptographic operations that are related to preparing JOSE structures. For the purposes of interoperability, DIDComm messaging does not support all JWAs, rather it takes a subset of the supported algorithms that are applicable for the following cases around secure messaging. These supported algorithms are listed here.
+
+
+### Sender Authenticated Encryption
+
+When a sender would like to encrypt a message for a recipient or multiple recipients and also be authenticated such that the recipients can determine they were the party that encrypted the message, the JWA of `ECDH-1PU` defined by [draft](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-02) SHOULD be used within the structure of a JWE. 
+
+For the keys involved in key agreement, the following elliptic curves MUST be supported.
+
+|Curve |Description                                                                                                                        |
+|------|-----------------------------------------------------------------------------------------------------------------------------------|
+|X25519|The underlying curve is actually `Curve25519`, however when used in the context of diffie helman the identifier of `X25519` is used|
+|P-256 |NIST defined P-256 elliptic curve                                                                                                  |
+
+For content encryption of the message, the following algorithms MUST be supported.
+
+|Algorithm(JWA) |Description                                          |
+|---------------|-----------------------------------------------------|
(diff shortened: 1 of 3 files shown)


encryption.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Sender Authenticated Encryption
+
+When a sender would like to encrypt a message for a recipient or multiple recipients and also be authenticated such that the recipients can determine they were the party that encrypted the message, the JWA of `ECDH-1PU` defined by [draft](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-02) SHOULD be used within the structure of a JWE. 
+
+For the keys involved in key agreement, the following elliptic curves MUST be supported.
+
+|Curve |Description                                                                                                                        |
+|------|-----------------------------------------------------------------------------------------------------------------------------------|
+|X25519|The underlying curve is actually `Curve25519`, however when used in the context of diffie helman the identifier of `X25519` is used|
```
kdenhartog, line 16: helman => hellman

encryption.md
```diff
@@ -0,0 +1,48 @@
...
+|P-256 |NIST defined P-256 elliptic curve                                                                                                  |
+
+For content encryption of the message, the following algorithms MUST be supported.
+
+|Algorithm(JWA) |Description                                          |
+|---------------|-----------------------------------------------------|
+|C20P           |ChaCha20Poly1305                                     |
```
kdenhartog, line 23: Use XC20P only. C20P means we have to manage nonces to keep it secure.

encryption.md
```diff
@@ -0,0 +1,48 @@
...
+|AES-GCM        |Advanced Encryption Standard with galois/counter mode|
```
kdenhartog, line 24: We can use random nonces for AES-GCM but need to be careful the number of messages sent. Practically speaking we shouldn't hit the limit (2^32 message), but need to state security considerations ( paging @mikelodder7 or @christianlundkvist ) around how nonces should be generated here or reference to an RFC which states how to do this safely (preferable).

encryption.md
```diff
@@ -0,0 +1,48 @@
...
+|Curve |Description                                                                                                                        |
+|------|-----------------------------------------------------------------------------------------------------------------------------------|
+|X25519|The underlying curve is actually `Curve25519`, however when used in the context of diffie helman the identifier of `X25519` is used|
+|P-256 |NIST defined P-256 elliptic curve                                                                                                  |
+
+For content encryption of the message, the following algorithms MUST be supported.
+
+|Algorithm(JWA) |Description                                          |
+|---------------|-----------------------------------------------------|
+|C20P           |ChaCha20Poly1305                                     |
```
kdenhartog, line 41: XC20P here with the definition being XChaCha20Poly1305

encryption.md
```diff
@@ -0,0 +1,48 @@
...
+|AES-GCM        |Advanced Encryption Standard with galois/counter mode|
```
kdenhartog, line 42: same as above about nonce reuse

encryption.md
kdenhartog: Of note because I'm sure it will be brought up Ideally, we would just use AES-GCM-SIV, but I don't believe that's NIST registered which is the reason we're supporting AES-GCM. For a nonce resistant cipher implementers should opt for XC20P if they don't have constraints on being NIST compliant. We should probably mention this as well.

OR13:
@dhh1128 related to your comment about different keys to different directions... https://libsodium.gitbook.io/doc/secret-key_cryptography/aead/chacha20-poly1305/ietf_chacha20-poly1305_construction#notes

Lots of good advice in libsodium docs.

kdenhartog:
> 
> 
> @dhh1128 related to your comment about different keys for different directions (sending /receiving)... https://libsodium.gitbook.io/doc/secret-key_cryptography/aead/chacha20-poly1305/ietf_chacha20-poly1305_construction#notes
> 
> Lots of good advice in libsodium docs.

Their recommendation is useful for regular chacha which supports only 96 bit nonces. If we use XChacha there's sufficient entropy that they can be stateless and random. See [here](https://tools.ietf.org/html/draft-irtf-cfrg-xchacha-03#section-2.1) for details.

encryption.md
TelegramSam: fixed.

encryption.md
TelegramSam: I've added a todo about nonces. Is that sufficient for this update?

encryption.md
TelegramSam: fixed.

encryption.md
TelegramSam: added a nonce TODO here as well. Does that work for now?

TelegramSam:
@kdenhartog How do these updates work?

awoie:
As per DIDComm WG call on 05/25/2020, we agreed to add explicit reference to IANA JOSE registry.

encryption.md
```diff
@@ -0,0 +1,52 @@
...
+### Sender Authenticated Encryption
+
+When a sender would like to encrypt a message for a recipient or multiple recipients and also be authenticated such that the recipients can determine they were the party that encrypted the message, the JWA of `ECDH-1PU` defined by [draft](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-02) SHOULD be used within the structure of a JWE. 
+
+For the keys involved in key agreement, the following elliptic curves MUST be supported.
+
+| Curve  | Description                                                  |
+| ------ | ------------------------------------------------------------ |
+| X25519 | The underlying curve is actually `Curve25519`, however when used in the context of Diffie-Hellman the identifier of `X25519` is used |
+| P-256  | NIST defined P-256 elliptic curve                            |
```
Baha-sk, line 17: Although `P-256` curve is the correct name, JWA also specifies the key wrapping algorithm using this curve. It's better to mention the algorithm value, especially that we're talking about ECDH here.

Since we're talking about ECDH-1PU encryption here, the kw algorithm would be: `ECDH-1PU+A256KW`. It's not part of the JWA spec yet, but it's defined in the ECDH-1PU draft link mentioned above this table.

Baha-sk: Probably adding a key agreement `algorithm` table will clarify this.

encryption.md
```diff
@@ -0,0 +1,52 @@
...
+### Anonymous Encryption
+
+When a sender would like to encrypt a message for a recipient or multiple recipients but not be authenticated by the recipients as the party who encrypted the message, the JWA of `ECDH-ES` defined by [RFC 7518](https://tools.ietf.org/html/rfc7518#section-4.6) SHOULD be used within the structure of a JWE. 
+
+For the keys involved in key agreement, the following elliptic curves MUST be supported.
+
+| Curve  | Description                                                  |
+| ------ | ------------------------------------------------------------ |
+| X25519 | The underlying curve is actually `Curve25519`, however when used in the context of Diffie-Hellman the identifier of `X25519` is used |
+| P-256  | NIST defined P-256 elliptic curve                            |
```
Baha-sk, line 37: Although `P-256` curve is the correct name, JWA also specifies the key wrapping algorithm using this curve. It's better to mention the algorithm value, especially that we're talking about ECDH here.

For instance, Anonymous encryption would be using ECDH-ES where the sender info is not required and a new ephemeral key is created at the send side for each new encryption. The ECDH key wrapping algorithm is called: `ECDH-ES+A256KW` (as per the [JWA spec](https://tools.ietf.org/html/rfc7518#section-4.6)).

As the key wrapping comment for ECDH-1PU above, it would help if we have a key wrapping algorithms table.

OR13: +1 for an algorithms table, cross linking: https://github.com/w3c-ccg/lds-jws2020/issues/12

We want to list the JsonWebKey recommended encryption `alg` here as well, and we need to ensure we support all the types that DIDComm uses.

TelegramSam: What form does this table show up in? I'm all for adding clarity here, but I have no idea how it ought to be properly represented in this document.
If you can help guide me, I'd like to get this done. I'm also open to accepting this PR as is and making the clarifications in another PR.
@Baha-sk @OR13

encryption.md
```diff
@@ -0,0 +1,52 @@
...
+
+For content encryption of the message, the following algorithms MUST be supported.
+
+| Algorithm(JWA) | Description                                           |
+| -------------- | ----------------------------------------------------- |
+| XC20P          | XChaCha20Poly1305                                     |
+| AES-GCM        | Advanced Encryption Standard with galois/counter mode |
```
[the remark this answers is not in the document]
TelegramSam: corrected as suggested.

encryption.md
```diff
@@ -0,0 +1,52 @@
...
```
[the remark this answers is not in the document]
TelegramSam: corrected as suggested

encryption.md
Baha-sk: I guess a new key wrapping algorithms table would look like this :
```
| KW Algorithm    | Curve (epk crv) | key type (epk kty) | Description                                                  |
| --------------- | --------------- | ------------------ | ------------------------------------------------------------ |
| ECDH-ES+A256KW  | P-256  | EC  | ECDH-ES key wrapping using key with NIST defined P-256 elliptic curve to create a 256 bits key as defined in [7518](https://tools.ietf.org/html/rfc7518#section-4.6.2)       |
| ECDH-ES+A256KW  | P-384  | EC  | ECDH-ES key wrapping using key with NIST defined P-384 elliptic curve to create a 256 bits key as defined in [7518](https://tools.ietf.org/html/rfc7518#section-4.6.2)       |
| ECDH-ES+A256KW  | P-521  | EC  | ECDH-ES key wrapping using key with NIST defined P-521 elliptic curve to create a 256 bits key as defined in [7518](https://tools.ietf.org/html/rfc7518#section-4.6.2)       |
| ECDH-ES+A256KW  | X25519 | OKP | ECDH-ES with X25519 ([RFC7748 section 5](https://tools.ietf.org/html/rfc7748#section-5)) to create a 256 bits key. The underlying curve is actually `Curve25519`, however when used in the context of Diffie-Hellman the identifier of `X25519` is used |
| ECDH-1PU+A256KW | P-256  | EC  | ECDH-1PU key wrapping using key with NIST defined P-256 elliptic curve to create a 256 bits key as defined in [ecdh-1pu](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03#section-2.2)      |
| ECDH-1PU+A256KW | P-384  | EC  | ECDH-1PU key wrapping using key with NIST defined P-384 elliptic curve to create a 256 bits key as defined in [ecdh-1pu](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03#section-2.2)      |
| ECDH-1PU+A256KW | P-521  | EC  | ECDH-1PU key wrapping using key with NIST defined P-521 elliptic curve to create a 256 bits key as defined in [ecdh-1pu](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03#section-2.2)      |
| ECDH-1PU+A256KW | X25519 | OKP | ECDH-1PU X25519 ([RFC7748 section 5](https://tools.ietf.org/html/rfc7748#section-5)) to create a 256 bits key as defined in [ecdh-1pu](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03#section-2.2)      |
```

how does this look?

encryption.md
```diff
@@ -0,0 +1,52 @@
+## Message Encryption
+
+DIDComm supports two types of message encryption: Authenticated Sender Encryption and Anonymous Sender Encryption. Both forms are encrypted to the recipient, but only Authenticated Sender Encryption provides assurances of who the sender is. 
+
+The encrypted form of a JWM is a JWE. The JOSE family defines [JSON Web Algorithms](https://tools.ietf.org/html/rfc7518) (JWAs) which standardize certain cryptographic operations that are related to preparing JOSE structures. For the purposes of interoperability, DIDComm messaging does not support all JWAs, rather it takes a subset of the supported algorithms that are applicable for the following cases around secure messaging. These supported algorithms are listed here.
+
+
+### Sender Authenticated Encryption
+
+When a sender would like to encrypt a message for a recipient or multiple recipients and also be authenticated such that the recipients can determine they were the party that encrypted the message, the JWA of `ECDH-1PU` defined by [draft](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-02) SHOULD be used within the structure of a JWE.
```
Baha-sk, line 10: draft 3 is out: https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03

encryption.md
TelegramSam: I'd feel more comfortable with a followup pr, submitted by somebody smarter about this stuff than I.

TelegramSam: Well, I got excited and included the table you included.

encryption.md
TelegramSam: updated

encryption.md
TelegramSam: added

TelegramSam:
@OR13 @Baha-sk @kdenhartog Any other comments before merging?

encryption.md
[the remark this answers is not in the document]
kdenhartog: Good catch, @Baha-sk

kdenhartog:
Looking good to me. I still have it as a follow up action item for myself to write the nonce portion of this. I've set out time next week to write them.

PR #56, opened 2020-05-19
per WG call, add comments about use of JSON-LD

dhh1128:


awoie:
As agreed per WG call on 05/25/2020 to merge this PR

PR #57, opened 2020-05-20
DID URI Type, Routing Processes

TelegramSam:
This PR contains (initially, anyway) two related topics: The process followed by both senders and mediators that supports routing, and the use of a DID alternative URI type for dereferencing mediator endpoints and keys.

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -282,8 +282,8 @@ These last two characteristics are the foundation of mix networking feature for 
 
 ### Sender Forward Process
 
-1. Sender Constructs Message
-2. Sender Encrypts Message to recipient(s)
+1. Sender Constructs Message.
+2. Sender Encrypts Message to recipient(s).
 3. Wrap Encrypted Message in Forward Message for each Routing Key.
 4. Transmit to `serviceEndpoint` in the manner specified in the [transports] section.
 
@@ -291,14 +291,14 @@ These last two characteristics are the foundation of mix networking feature for 
 
 _Prior to using a Mediator, it is the recipient's responsibility to coordinate with the mediator. Part of this coordination informs them of the `to` address(es) expected, the endpoint, and any Routing Keys to be used when forwarding messages. That coordination is out of the scope of this spec._
 
-1. Receives Forward Message
-2. Retrieves Service Endpoint and Routing Keys as pre-configured by recipient (`to` attribute).
-3. Wrap attached `payload` message once per Routing Key
+1. Receives Forward Message.
+2. Retrieves Service Endpoint and (optionally) Routing Keys as pre-configured by recipient (`to` attribute).
+3. If Routing Keys pre-configured (step 2), wrap attached `payload` message once per Routing Key.
 4. Transmit to Service Endpoint in the manner specified in the [transports] section.
 
 ### DID Document Keys
 
-All keys declared in the DID Document's `keyAgreement` section should be used to encrypt messages. The details of key representation are described in the [Public Keys section of the DID Core Spec](https://www.w3.org/TR/did-core/#public-keys).
+All keys declared in the DID Document's `keyAgreement` section should be used as recipients when encrypting a message. The details of key representation are described in the [Public Keys section of the DID Core Spec](https://www.w3.org/TR/did-core/#public-keys).
 
 Keys used in a signed JWM are declared in the DID Document's `authentication` section.
 
@@ -321,13 +321,13 @@ DIDComm DID Document endpoints have the following format:
 
 **type**: MUST be `DIDComm`. 
 
-**serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec, or a URI from Alternative Endpoints. Endpoints from the [transports] section SHOULD be used only for the reception of DIDComm messages.
+**serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec, or a URI from Alternative Endpoints. It MAY be desirable to constraint endpoints from the [transports] section so that they are used only for the reception of DIDComm messages. This can be particularly helpful in cases where auto-detecting message types is inefficient or undesirable.
 
 **routingKeys**: An optional ordered array of strings referencing keys to be used when preparing the message for transmission as specified in the [Routing] section of this spec. 
 
 #### Multiple Endpoints
 
-A DID Document may contain multiple service entries of type `DIDComm`. Unless otherwise specified, messages should be sent to endpoints in the order they are specified.
+A DID Document may contain multiple service entries of type `DIDComm`. Entries are SHOULD be specified in order of receiver preference, but any endpoint MAY be selected by the sender, typically by protocol availability or preference.
 
 #### Alternative Endpoints
 
--- a/routing.md
+++ b/routing.md
@@ -295,7 +295,7 @@ _Prior to using a Mediator, it is the recipient's responsibility to coordinate w
 2. Retrieves Service Endpoint pre-configured by recipient (`to` attribute).
 4. Transmit `payload` message to Service Endpoint in the manner specified in the [transports] section.
 
-The recipient (`to` attribute of Forward Message) may have pre-configured additional routing keys with the mediator. If this is the case, the mediator should wrap the attached `payload` message into an additional Forward message once per routing key. This step is performed between (2) and (3).
+The recipient (`to` attribute of Forward Message) may have pre-configured additional routing keys with the mediator that were not present in the DID Document and therefore unknown to the original sender. If this is the case, the mediator should wrap the attached `payload` message into an additional Forward message once per routing key. This step is performed between (2) and (3).
 
 ### DID Document Keys
 
--- a/routing.md
+++ b/routing.md
@@ -337,7 +337,7 @@ In addition to the URIs for [transports], some alternative forms are available.
 
 Using a DID for the `serviceEndpoint` is useful when using a mediator. The DID should be resolved, and services with type of "DIDComm" will contain valid `serviceEndpoints`. The keyAgreement keys of that DID Document should be appended at the end of the routingKeys section from the message recipient's DID Document as per the process in [Sender Forward Process]. The key advantage with this approach is that a mediator can rotate keys and update serviceEndpoints without any updates needed to dependent recipients` DID Documents.
 
-A DID used as a mediator in this way SHOULD NOT use alternative endpoints. Only URIs described in [transports] are should be used to avoid recursive lookups.
+A DID representing a mediator SHOULD NOT use alternative endpoints in it's own DID Document to avoid recursive endpoint resolution. Using only the URIs described in [transports] will prevent such recursion.
 
 Example 1: Mediator
 
--- a/routing.md
+++ b/routing.md
@@ -292,9 +292,10 @@ These last two characteristics are the foundation of mix networking feature for 
 _Prior to using a Mediator, it is the recipient's responsibility to coordinate with the mediator. Part of this coordination informs them of the `to` address(es) expected, the endpoint, and any Routing Keys to be used when forwarding messages. That coordination is out of the scope of this spec._
 
 1. Receives Forward Message.
-2. Retrieves Service Endpoint and (optionally) Routing Keys as pre-configured by recipient (`to` attribute).
-3. If Routing Keys pre-configured (step 2), wrap attached `payload` message once per Routing Key.
-4. Transmit to Service Endpoint in the manner specified in the [transports] section.
+2. Retrieves Service Endpoint pre-configured by recipient (`to` attribute).
+4. Transmit `payload` message to Service Endpoint in the manner specified in the [transports] section.
+
+The recipient (`to` attribute of Forward Message) may have pre-configured additional routing keys with the mediator. If this is the case, the mediator should wrap the attached `payload` message into an additional Forward message once per routing key. This step is performed between (2) and (3).
 
 ### DID Document Keys
 
--- a/routing.md
+++ b/routing.md
@@ -321,7 +321,7 @@ DIDComm DID Document endpoints have the following format:
 
 **type**: MUST be `DIDComm`. 
 
-**serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec, or a URI from Alternative Endpoints.
+**serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec, or a URI from Alternative Endpoints. Endpoints from the [transports] section SHOULD be used only for the reception of DIDComm messages.
 
 **routingKeys**: An optional ordered array of strings referencing keys to be used when preparing the message for transmission as specified in the [Routing] section of this spec. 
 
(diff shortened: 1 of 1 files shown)


routing.md
```diff
@@ -305,12 +321,51 @@ DIDComm DID Document endpoints have the following format:
...
+
+A DID used as a mediator in this way MUST NOT use alternative endpoints. Only URIs described in [transports] are acceptable.
+
+Example 1: Mediator
+
+```json
+{
+    "id": "did:example:123456789abcdefghi#didcomm-1",
+    "type": "DIDComm",
+    "serviceEndpoint": "did:example:somemediator"
```
kdenhartog, line 348: This will need to be a didUrl rather than a did, so the specific service endpoint can be de-referenced properly.

routing.md
```diff
@@ -305,12 +321,51 @@ DIDComm DID Document endpoints have the following format:
...
+```
+The message is encrypted to the recipient, then wrapped in a forward message encrypted to the keyAgreement keys within the `did:example:somemediator` DID Document, and transmitted to the URIs present in the `did:example:somemediator` DID Document with type `DIDComm`.
+
+Example 2: Mediator + Routing Keys
+```json
+{
+    "id": "did:example:123456789abcdefghi#didcomm-1",
+    "type": "DIDComm",
+    "serviceEndpoint": "did:example:somemediator",
+    "routingKeys": ["did:example:anothermediator#somekey"]
```
kdenhartog, line 359: As we discussed awhile back, I don't believe this is going to be a usable method of specifying a route. Putting this list of keys in the did document, where the route will likely need to be changing quite frequently (meaning keys will be updated frequently) will almost certainly cause us a large number of headaches. I'd suggest we find a different method other than this. I personally think the JIT routing scheme would be a better approach because we would not need to put routing keys in the did document.

routing.md
TelegramSam: does it? What if any of the transports are valid? As the goal is to get the message to them, I don't think that locking it to one specific transport is helpful.

routing.md
TelegramSam: We do need a discussion of this, but I wanted to note something important: The list of routingKeys is NOT the complete list of keys in the route. Any number of additional routing steps can exist in between and after the keys listed here. The keys listed here are ONLY the keys shared to the sender from the recipient. 

Use of these keys allow hiding the header information of the final message from mediators on the route, a feature that I don't believe pure JIT routing (only one mediator in the DID Doc) allows.

kdenhartog: > 
> 
> We do need a discussion of this, but I wanted to note something important: The list of routingKeys is NOT necessarily the complete list of keys in the route. Any number of additional routing steps can exist in between and after the keys listed here. The keys listed here are ONLY the keys shared to the sender from the recipient.
> 
> Use of these keys allow hiding the header information of the final message from mediators on the route, a feature that I don't believe pure JIT routing (only one mediator in the DID Doc) allows.

Which header information are we hiding? The only thing I can think of is the recipient information. Onion routing doesn't solve this completely either as the last node (referred to as the exit node in TOR) is still able to identify the recipient. I'd like us to really assess what the tradeoffs are here. Using onion routing feels like an extension use case rather than the default requirement and I don't think we should be intentionally designing to meet it's needs. Rather we should design for simplicity and allow the people who want to extend to support this functionality to define how that should be done.

routing.md
kdenhartog: serviceEndpoint would need to be a URL is what I'm saying. did:example:123 is a URI did:example:123#endpoint1 is a URL.

See [here](https://w3c.github.io/did-core/#did-syntax) for the URI syntax and [here](https://w3c.github.io/did-core/#did-url-syntax) for the did URL syntax

TelegramSam: The DID Core Spec indicates that it should be a URI. We could further constrain that for the DIDComm use case, but allowing a DID (URI) has a very useful application here and is allowable.

So, I disagree that serviceEndpoint needs to be a URL.

routing.md
TelegramSam: The recipient information is precisely what we are trying to hide, from all but the last step.

And yes, we should discuss the tradeoffs. We may not be able to complete that conversation in today's meeting, as the US holiday may not allow attandance of key folks for the conversation.

routing.md
kdenhartog: Agree now that I understand how it's being used. We'll need to further define expected behavior of uri resolution. (E.g. MUST select a service endpoint from the resolved did doc with type: DIDComm, SHOULD NOT resolve deeper then two redirects)

routing.md
kdenhartog: Ok, let's dig a bit deeper on this here while we can. Who MUST we hide the recipient info from and/or who SHOULD we hide the recipient info from? Let's start there to keep the conversation constrained (you know how much I love to squirrel in topics like this).

kdenhartog: @dhh1128 @tplooker @swcurran you'll probably want to be apart of this conversation.

dhh1128: It seems like we are thinking of JIT and onion routing as opposites. That's a useful simplification, sometimes -- but it's a distinction that breaks down under various use cases. Of course we can have the simple case where Bob sends to Alice via her agency and via her cloud mediator. We've discussed that at length. But there is nothing to prevent far more complex scenarios. Suppose Alice wants to advertise her public mediator to be AgentsRUs's DID, which has separate instructions to forward to her cloud agent, which which does not appear in her `routingKeys` array at all. In such a case, the cloud agent is incapable of reading a message (even a forward) -- but perhaps it is configured as a relay, which in turn re-encrypts (NOT proxy re-encrypts) and sends the message on to a different DID hosted at a different agency. Repeat 10x. Eventually the message ends up at an agent (NOT a cloud agent that's playing the mediator role) that can directly decrypt the message. This is quite a bit like constructing a channel that consists of interlocking SSH tunnels, where parts of the route (the first one, specifically) use onion routing, and other parts use JIT.

In such a scenario, what does the sender (Bob) need to know about Alice's routing? Just her first public mediator. To him it looks like onion routing, but that's not the whole story, and that's the point. We are NOT representing all the hops in a route with this array. We are representing exactly and only as many hops as the sender, Bob, needs to know to make delivery work. Alice can change the other details to suit her preference.

What will not change is that the final mediator (who unwraps the innermost `forward` message) MUST see the recipients, and all other hops MUST NOT see it unless Alice (not Bob) chooses to expose it.

swcurran: What @dhh1128 said so well.  The receiver tells the sender (via the DIDComm service) to do certain wrappings.  What actually happens in the flow is a separate issue other than the messages must be delivered to the endpoint and at some point be decrypted in order.

TelegramSam: I love everything about the last few comments here. What is needed to capture this? What can I add to the PR to make this mergable?

routing.md
TelegramSam: Line 340 already contains the MUST NOT recurse (use of alternative endpoints).
Line 351 already contains the requirement for the mediator endpoint to be of type `DIDComm`.

Does the language of either need to be adjusted?

TelegramSam: I did adjust line 340 to use SHOULD NOT instead of MUST NOT.

TelegramSam:
I added a note about endpoint exclusivity that came up in a discussion about discovering DIDComm types.

routing.md
```diff
@@ -280,11 +280,27 @@ Why is such indirection useful?
...
+3. Wrap Encrypted Message in Forward Message for each Routing Key.
+4. Transmit to `serviceEndpoint` in the manner specified in the [transports] section.
+
+### Mediator Process
+
+_Prior to using a Mediator, it is the recipient's responsibility to coordinate with the mediator. Part of this coordination informs them of the `to` address(es) expected, the endpoint, and any Routing Keys to be used when forwarding messages. That coordination is out of the scope of this spec._
+
+1. Receives Forward Message
+2. Retrieves Service Endpoint and Routing Keys as pre-configured by recipient (`to` attribute).
+3. Wrap attached `payload` message once per Routing Key
```
dhh1128, line 296: I don't think a mediator necessarily performs wrapping. They may (and I think in most cases will) receive a forward message, unwrap it, and send what they unwrapped along to the next set of recipients, without performing a re-wrapping.

routing.md
```diff
@@ -280,11 +280,27 @@ Why is such indirection useful?
...
+4. Transmit to Service Endpoint in the manner specified in the [transports] section.
+
 ### DID Document Keys
 
 All keys declared in the DID Document's `keyAgreement` section should be used to encrypt messages. The details of key representation are described in the [Public Keys section of the DID Core Spec](https://www.w3.org/TR/did-core/#public-keys).
```
dhh1128, line 301: suggest a tweak: "should be used to encrypt messages" --> "should be used as recipients when encrypting messages"

routing.md
```diff
@@ -305,12 +321,51 @@ DIDComm DID Document endpoints have the following format:
 
 **type**: MUST be `DIDComm`. 
 
-**serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec.
+**serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec, or a URI from Alternative Endpoints. Endpoints from the [transports] section SHOULD be used only for the reception of DIDComm messages.
```
dhh1128, line 324: The sentence "Endpoints from the [transports] section SHOULD..." is good advice for some transports. This includes http, which is probably why it feels right to us. But if the transport = email, I'm not sure the sentence makes sense. If the transport is bluetooth, I'm not sure it does, either. I guess I'm challenging the notion that an endpoint that accepts DIDComm should only accept DIDComm. How would we feel about rewording to "It MAY be desirable to constraint endpoints from the [transports] section so that they are used only for the reception of DIDComm messages. This can be particularly helpful in cases where auto-detecting message types is inefficient or undesirable."

routing.md
```diff
@@ -305,12 +321,51 @@ DIDComm DID Document endpoints have the following format:
...
 
 **routingKeys**: An optional ordered array of strings referencing keys to be used when preparing the message for transmission as specified in the [Routing] section of this spec. 
 
 #### Multiple Endpoints
 
 A DID Document may contain multiple service entries of type `DIDComm`. Unless otherwise specified, messages should be sent to endpoints in the order they are specified.
```
dhh1128, line 330: I don't think this "in the order specified" is quite true. The first criterion ought to be a compatible transport for the circumstances; order of endpoints in the doc would be secondary. Suppose I have 5 DIDComm endpoints: 1= http, 2=http, 3=chat, 4=smtp, 5=bluetooth. If I'm disconnected from the internet, only bluetooth is useful, so I pick that. If I'm connected, I might pick any of endpoints 1-4, depending on whether I'm streaming large chunks of data (http), using the BasicMessage protocol (chat or smtp), etc. If I know I'm using http, I should prefer 1 over 2.

routing.md
dhh1128: BTW, I'm ambivalent about my previous comment. I'm effectively thinking out loud, not making a strong assertion.

routing.md
TelegramSam: I'll see if I can adjust the language to represent the optionality of this.

routing.md
TelegramSam: I like you reword.

routing.md
TelegramSam: I've updated the language to address your points.

TelegramSam:
@dhh1128 thanks for the comments. I think I've addressed them all.

routing.md
```diff
@@ -282,23 +282,23 @@ These last two characteristics are the foundation of mix networking feature for
...
 ### Mediator Process
 
 _Prior to using a Mediator, it is the recipient's responsibility to coordinate with the mediator. Part of this coordination informs them of the `to` address(es) expected, the endpoint, and any Routing Keys to be used when forwarding messages. That coordination is out of the scope of this spec._
 
-1. Receives Forward Message
-2. Retrieves Service Endpoint and Routing Keys as pre-configured by recipient (`to` attribute).
-3. Wrap attached `payload` message once per Routing Key
+1. Receives Forward Message.
+2. Retrieves Service Endpoint and (optionally) Routing Keys as pre-configured by recipient (`to` attribute).
+3. If Routing Keys pre-configured (step 2), wrap attached `payload` message once per Routing Key.
```
dhh1128, line 296: This still does not match my mental model. Mediators do not commonly wrap at all. They unwrap (decrypt a `forward` message) and then forward, without rewrapping, the payload inside it. They do this behavior even if routing keys are pre-configured. The only time rewrapping would be done is if they are sending to a mediator that's not visible to the outside world.

TelegramSam: The ability for the recipient to arrange routing not visible to the sender is one of the useful things here. In simplest form a mediator just knows which endpoint to use for the internal payload. We could eliminate this option in favor of only the last mediator being capable of doing something like this, but I was generalizing the pattern a little. 

Are you in favor of eliminating this for the sake of simplicity?

dhh1128: Not at all. I don't see why the two ideas are connected. Mediators don't have to always re-wrap to ensure this flexibility. They only re-wrap when the flexibility has been activated; they don't do it in all cases. The sender doesn't need to know which behavior they're using.

routing.md
```diff
@@ -305,12 +321,51 @@ DIDComm DID Document endpoints have the following format:
...
+
+#### Alternative Endpoints
+
+In addition to the URIs for [transports], some alternative forms are available.
+
+##### DID
+
+Using a DID for the `serviceEndpoint` is useful when using a mediator. The DID should be resolved, and services with type of "DIDComm" will contain valid `serviceEndpoints`. The keyAgreement keys of that DID Document should be appended at the end of the routingKeys section from the message recipient's DID Document as per the process in [Sender Forward Process]. The key advantage with this approach is that a mediator can rotate keys and update serviceEndpoints without any updates needed to dependent recipients` DID Documents.
+
+A DID used as a mediator in this way SHOULD NOT use alternative endpoints. Only URIs described in [transports] are should be used to avoid recursive lookups.
```
kdenhartog, line 340: I'm not sure I followed this statement. The way I read and understood it was "A mediator represents their did document in particular ways. One of the ways it SHOULD NOT use it though is to redirect to a different endpoint if it knows it's a mediator service. To avoid this problem mediator services should define their service endpoints in a way that are described in the transports section".

1. Am I understanding this correctly?
2. If so, I'd suggest rewording below:

```suggestion
A DID Document known to represent a mediator service SHOULD NOT use a redirecting endpoint to avoid recursive loops. Rather, it should directly represent it's service endpoints as URIs described in the [transports] section. Additionally, to avoid recursion loops, it's RECOMMENDED that mediator services not configure DID Document recursion layers deeper than 2.
```

routing.md
kdenhartog: Thanks, I made one minor suggestion to the wording because I wasn't certain I understood what you were trying to say. In general, I think in general, we've addressed my concerns though and anything additional is editorial remarks.

routing.md
TelegramSam: I see the confusion you point out, but I was also confused a bit by your suggestion. I've reworded this section; please review and see if it satisfies the intent of your comment.

routing.md
TelegramSam: I've updated the language of this section to extract the optional processing of additional routing keys below the process steps. Does this work for you?

dhh1128: No. The fundamental misalignment in mental models is still important: mediators usually don't wrap. Let's discuss interactively.

PR #69, opened 2020-06-29
add message authorization section

kdenhartog:
This adds details about how the from field should be associated with the keys and kid used in the JWE/JWS so the message can be authorized.


jwm.md
```diff
@@ -94,4 +94,8 @@ For digital signatures the following algorithms MUST be supported.
 
 ## Signed and Encrypted
 
-When a sender would like for their message to feature both a non-repudiable digital signature and be encrypted to a recipient or multiple recipients. They can combine the encryption and signing operations defined earlier in this section by creating a [nested JWM](https://tools.ietf.org/html/draft-looker-jwm-01#section-1.2).
\ No newline at end of file
+When a sender would like for their message to feature both a non-repudiable digital signature and be encrypted to a recipient or multiple recipients. They can combine the encryption and signing operations defined earlier in this section by creating a [nested JWM](https://tools.ietf.org/html/draft-looker-jwm-01#section-1.2).
+
+## Message Authorization
+
```
TelegramSam, line 100: What does it mean for a message to be authorized?
Context for evaluation based on time, or version on the DID's Document?
[Notes from WG meeting]

jwm.md
```diff
@@ -94,4 +94,8 @@ For digital signatures the following algorithms MUST be supported.
...
...ST validate that the keys used by the sender in the JWE and/or JWS are valid and within the did document of the expected decentralized identifier. To verify the authorization of the message, the recipient MUST resolve a valid DID Document associated with the DID in the `from` field of the JWM message. The recipient MUST then validate that the keys and key ids (kid) used in the JWE/JWS match keys and identifiers associated (via the did document) with the decentralized identifier. The recipient also MUST verify that a key used for [Sender Authenticated Encryption](#Sender-Authenticated-Encryption) is associated with the verification relationship property named `keyAgreement` in order for the message to be authorized. In the case of a [signed](#Signing) message, the key must maintain a verification relationship with the `assertionMethod` property in order for the message to be authorized. In the case of a [nested](Signed-and-Encrypted) message the associated keys and key ids MUST be valid for the message to be considered authorized. If the keys are not contained within the did document then the message should not be considered authorized, even if the cryptographic operations succeeds.
```
OR13, line 101: `assertionMethod` is for creating VCs... I would expect `authentication` in this case.

OR13:
I recall discussing this on the call, i believe the action was to split up the section, so we can clearly see what we are trying to accomplish case by case.

kdenhartog:
Got this fixed up to include the feedback now @OR13 and @TelegramSam

@dhh1128 please take a look at the note as well to make sure the language satisfies the edge case of returning invalid because the DID Document has been updated.

jwm.md
kdenhartog: changed to authentication and added a note at the bottom to cover this.

TelegramSam:
This looks good to me.

dhh1128:
I'm cool with a merge.

Issue #33, opened 2020-02-28
Should we represent service block info as a DID with `initial-state` matrix params?


TelegramSam:
Per discussion on Slack, it has been suggested that we leverage the `initial-state` matrix parameter to represent service block info when unattached to a DID. This is common both in the invitation stage of exchanging DIDs and with ephemeral challenges, where endpoint, encryption keys, and routing keys are passed.

Each DID method can define how `initial-state` values apply to the method. `did:peer` could use this, and we might also work this into `did:key` to attach the additional needed information. As the method defines use, we can design this to be concise.

Thoughts:
- `initial-state` feels young, even to the point of renaming what the name will be. Matrix parameters are also under active discussion. Using the same approach may provide external pain as we adapt to changing external decisions.
- Using this approach means that the value is always a DID and may be passed to a library for DID method based evaluation. This provides a consistent dev experience between resolved DIDs and information passed inline (like did:key+service info).

dhh1128:
I'm just leaving a comment so I'm subscribed to this issue. (Clicking "subscribe" in github hasn't been working for me.)

OR13:
Can someone with more experience using `did:peer` describe in concrete terms what the KERI inception event like thing is that happens when a `did:peer` is first created?

I have a feeling understanding that is the key to this issue.

In sidetree the `initial-state` is an unwitnessed inception event, which is acceptable only when no witnessed inception event is observable on the ledger.

csuwildcat:
> In sidetree the `initial-state` is an unwitnessed inception event, which is acceptable only when no witnessed inception event is observable on the ledger.

I like how you put this, that's a great way to word it!

OR13:
@SmithSamuelM  ^ We have a couple conversations floating around related to this, which I would love your thoughts on.

dhh1128:
I'll answer Orie's question for peer DIDs specifically. Then I'll make some observations about principles of self-certifying identifiers that @smithsamuelm has been talking about. Sam's principles are embodied in did:peer, but not as purely as the way Sam advocates (peer DIDs mix in a couple other concerns). Sam has done a beautiful job of abstracting some ideas further than peer DID does.

[...]

PR #73, opened 2020-07-02
Removal of OOB and DID Exchange information

TelegramSam:
Given the move to zero round trip models, these items are not longer needed.


tplooker:
LGTM, will await others review before deciding to merge

connections.md
```diff
@@ -2,254 +2,7 @@
...
 A Connection is the practical application of a relationship in DID Communication. Having a connection means that each party in the relationship has a DID for the other parties, and parties can communicate securely using the keys and endpoints within each DID Document.
 
-In order to establish a new connection, we use the OutOfBand Protocol and the DID Exchange Protocol.
+In order to establish a new connection, Simply exchange a new message between parties. Knowing the DID of the other parties does not indicate any level of trust. Using the connection to establish a foundation of trust is the next step.
 
 
 
-#### OutOfBand Protocol
-
-Invitations are usually the first message passed between parties. They are passed out of band, such as in a QR code or a link that can either open in an app or in a browser containing getting started instructions.
```
troyronda, line 11: Are the concepts in OOB invitations being moved elsewhere in the spec?

connections.md
```diff
@@ -2,254 +2,7 @@
 
 A Connection is the practical application of a relationship in DID Communication. Having a connection means that each party in the relationship has a DID for the other parties, and parties can communicate securely using the keys and endpoints within each DID Document.
 
-In order to establish a new connection, we use the OutOfBand Protocol and the DID Exchange Protocol.
+In order to establish a new connection, Simply exchange a new message between parties. Knowing the DID of the other parties does not indicate any level of trust. Using the connection to establish a foundation of trust is the next step.
```
troyronda, line 5: > Using the connection to establish a foundation of trust is the next step.

Are any aspects of this next step part of the DIDComm spec?

troyronda, line 5: > In order to establish a new connection

At what point is the connection considered "established" by each party? Is it when the party *receives* a message back from the other party? (does the act of sending a message to a new "DID" constitute a pending connection until a message is returned?)

connections.md
```diff
@@ -2,254 +2,7 @@
...
-
-Some interactions do not need an established connection to complete their purpose. These interactions occur in _ephemeral mode_ and use OutOfBand protocol messages for the bulk of the interaction.
-
-TODO: Include more details about ephemeral mode.
-
-#### DID Exchange Protocol
-
-##### 1. Exchange Request
-
-The exchange request message is used to communicate the DID document of the _invitee_ to the _inviter_ using the provisional service information present in the _invitation_ message.
```
troyronda, line 29: I assume this purpose of the DID Exchange Protocol is obsoleted by usage of initial-state.

I am wondering about the nuance of peer DIDs and initial-state after the connection is considered established. Should initial-state continue to be sent after that point?

connections.md
```diff
@@ -2,254 +2,7 @@
...
-
-TODO: Add clear details about initial_state usage?
-
-Example
-
-```json
-{
-  "@id": "5678876542345",
-  "@type": "https://didcomm.org/didexchange/2.0/request",
-  "~thread": { "pthid": "<id of invitation>" },
```
troyronda, line 43: The correlation to invitations is a useful concept.

llorllale:
I would first like to see the alternative in detail before actually removing this content.

connections.md
troyronda, line 5: I am also curious about scenarios where a sender wants an established connection _prior_ to some action/message/protocol. In some cases, they may wish to confirm that the recipient is able to decrypt their messages; that the route is working; and that the recipient can respond with authenticated messages; (and authenticate the DID). For example, during issuance or (generally) authentication scenarios.

connections.md
TelegramSam: Sort of. The ability to pass a message as an attachment isn't needed under the proposed 0RTT model.

connections.md
TelegramSam: This is a great question. I've been spending some time considering our mental models around connections. More on this to come.

connections.md
TelegramSam: I think yes, though I think initial-state might not be the right term for that.

connections.md
TelegramSam: I think that decision is up to us - which components are so core that they need to be in the spec, and which are better left to higher layers? I consider this to be very unresolved as yet.

connections.md
troyronda: The OutOfBand protocol provides a common mechanism to correlate a "not DIDComm activity" to a DIDComm message. This concept isn't about 0RTT vs DID Exchange but rather about defining a common mechanism for this capability.

connections.md
```diff
@@ -2,254 +2,7 @@
...
-
-The _inviter_ should check the information presented with the keys used in the wire-level message transmission to ensure they match.
-
-The _inviter_ MAY look up the corresponding invitation identified in the request's `~thread.pthid` to determine whether it should accept this exchange request.
-
-If the _inviter_ wishes to continue the exchange, they will persist the received information in their wallet. They will then either update the provisional service information to rotate the key, or provision a new DID entirely. The choice here will depend on the nature of the DID used in the invitation.
-
-The _inviter_ will then craft an exchange response using the newly updated or provisioned information.
-
-###### Request Errors
```
troyronda, line 119: I am curious what common error messages will remain in the spec.

troyronda, line 119: I am curious about the difference in errors originating from OOB/DID Exchange vs the DIDComm v2 model?

connections.md
troyronda: Scenario: Alice has received an “invitation” (not in a DIDComm message) from Bob which then triggers Alice to send a DIDComm message to Bob.

OutOfBand defines an invitation format, goal/goal code, versioning information, and enables correlation based on the invitation message ID to a subsequent DIDComm message thread.

connections.md
troyronda: If the parties have an implicit connection establishment model (on top of the 0RTT any message model), then perhaps they could choose to stop sending initial-state once established. Without knowledge that the other party has persisted then, yes, I guess they would have to include with every message.

connections.md
TelegramSam: By (not in a DIDComm message) I think you mean "Not in a DIDComm Encrypted message", but still in one of the standard formats in the spec? (example: URL/QR code, or another yet to be defined format?)

Invitation format: See the QR code PR for defined format. #74 
Versioning Information: Only pertained to the handshake protocols, which are not needed given the presence of the `from` and `from_prior` headers.
Correlation based on message ID: Correlate with the message ID of the message sent.

Goal/Goal Code: There isn't currently a replacement for this, or actually for the 'I want to exchange DIDs for no declared reason'. I believe the answer to this is deeper and related to the discussion about the ux of a user accepting an invitation. I'm going to open an issue to discuss this particular topic.

connections.md
TelegramSam: More thoughts on this, after discussion in both the Aries WG call and the DIDComm WG call today.

The historical context of 'established' is very binary: a connection is either 'established' or it isn't.  I believe the right answer is closer to this: When you receive a message from the other party, you know their information about how to contact you is accurate. When they reply to something you sent, you know your information about how to contact them is accurate.

connections.md
TelegramSam: Agree. How the further changes to a did:peer did are communicated to the various parties is something that the method will need to define. Perhaps deltas a communicated in a query param as well, and dropped upon confirmation of receipt.

connections.md
TelegramSam: answered in above comment. tldr; You can still correlate to the message ID that is passed.

connections.md
TelegramSam: Can you pick an example to discuss?

troyronda: If I were to generalize the thought: How do we differentiate between a protocol-layer errors vs lower-layer concerns that were previously in these pre-step connection protocols. As an example, an unsupported DID method happened to be used (or unsupported key type or unsupported endpoint, etc).

Of course there is also possibilities such as unsupported protocols (or versions), but these would be cases that could happen regardless.

TelegramSam: I don't believe OOB has anything in the spec about any of the errors you mentioned. Those are certainly important to think about, but I don't think there is a difference there between OOB/DIDExchange and 0RTT.

woiue:
@TelegramSam on the last didcomm wg call, people agreed to put OOB in its own section. People were also more in favour of naming `_dc` `_oob`. In terms of naming, while for us it is quite obvious we should also define what `_dc` or `_oob` stands for.

TelegramSam:
Updated all those items in the other PR.

connections.md
llorllale: I think we've lost a cheap and sensible way to create an OOB invitation with explicit didcomm parameters (eg. no public DID).... see my comments here: https://github.com/decentralized-identity/didcomm-messaging/issues/33#issuecomment-680052846

connections.md
llorllale: @TelegramSam can you point to the verbiage that describes the field used to indicate the thread's ID?

llorllale: It's not in the [DIDComm JWM Profile](https://github.com/decentralized-identity/didcomm-messaging/blob/master/jwm.md#jwm-profile)

connections.md
```diff
@@ -2,254 +2,7 @@
...
-}
-```
-
-###### Attributes
-
-* The `@type` attribute is a required string value that denotes that the received message is an exchange request.
-* The [`~thread`](../../concepts/0008-message-id-and-threading/README.md#thread-object) decorator MUST be included:
-  * It MUST include the ID of the parent thread (`pthid`) such that the `request` can be correlated to the corresponding `invitation`. More on correlation [below](#correlating-requests-to-invitations).
-  * It MAY include the `thid` property. In doing so, implementations MUST set its value to that of `@id` on the same request message. In other words, the values of `@id` and `~thread.thid` MUST be equal.
-* The `label` attribute provides a suggested label for the DID being exchanged. This allows the user to tell multiple exchange requests apart. This is not a trusted attribute.
```
llorllale, line 55: We're losing the `label` feature... what's the replacement?

dhh1128:
+1 to merge from me

connections.md
TelegramSam: It seems that was missing from the OOB PR when it was approved. I think the best replacement of the label in an OOB invitation would be in the form of a PR to `out_of_band.md`.

llorllale: Discussed on today's call: we need to push in a PR to the OOB text to add the label

troyronda:
For those that want to validate a connection is _established_ prior to another action, we will still need a protocol to do so.

Related: https://github.com/decentralized-identity/didcomm-messaging/issues/81

troyronda:
Regarding the usage of a ping protocol to enable an explicit RT at connection establishment.

In addition to using an authenticated RT ping, the other aspect is to enable linking that ping to an OOB request (e.g., stating the parent thread). This enables the Party to (A) know to respond to the ping; and (B) tie together the authenticated messages to an initial OOB action.

llorllale:
> Regarding the usage of a ping protocol to enable an explicit RT at connection establishment.
> 
> In addition to using an authenticated RT ping, the other aspect is to enable linking that ping to an OOB request (e.g., stating the parent thread). This enables the Party to (A) know to respond to the ping; and (B) tie together the authenticated messages to an initial OOB action.

See #95

connections.md
```diff
@@ -2,254 +2,7 @@
...
-  "label": "Bob",
-  "did": "<did invitee is sending to inviter>"
-}
-```
-
-###### Attributes
-
-* The `@type` attribute is a required string value that denotes that the received message is an exchange request.
-* The [`~thread`](../../concepts/0008-message-id-and-threading/README.md#thread-object) decorator MUST be included:
-  * It MUST include the ID of the parent thread (`pthid`) such that the `request` can be correlated to the corresponding `invitation`. More on correlation [below](#correlating-requests-to-invitations).
```
troyronda, line 53: parent thread will need to be defined elsewhere in the spec.

troyronda: https://github.com/decentralized-identity/didcomm-messaging/issues/95#issuecomment-692226190

connections.md
llorllale: Submitted PR #102 to add threading

tplooker (approved):

PR #88, opened 2020-08-24
Add guide link for rendered guide.

TelegramSam:


dhh1128:
Quite uncontroversial, I think. Discussed on WG call.

PR #90, opened 2020-08-24
Return-route extension

TelegramSam:

### DIFF
--- a/extensions/return_route/main.md
+++ b/extensions/return_route/main.md
@@ -1,0 +1,27 @@
+# DIDComm Messaging Return-Route Extension
+
+## Summary
+
+DIDComm Transports are simplex: they only transfer messages from sender to receiver. This extension enables bi-directional communication on the same transports. Messages can flow back in response to inbound messages over the same connection. This is particularly useful for communication between agents that are unable to provide routable endpoints (such as mobile phones or agents inside a firewall) and their mediators.
+
+## Scope
+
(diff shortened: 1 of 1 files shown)


extensions/return_route/main.md
```diff
@@ -0,0 +1,27 @@
...
+
+This extension is only to be used between the immediate sending agent and the agent it connects directly to. This extension is not valid when routing keys are in use.
+
+## Reference
+
+This extension adds a new header in DIDComm messages: `return_route`. When a message is received with this header, use of the connection to return messages should be adjusted according to the value presented.
+
+- `return_route` has the following acceptable values:
+
+  - `none`: Default. No messages should be returned over this connection.
```
kdenhartog, line 17: Could we optimize this by saying if it's not included it can be assumed that No messages should be returned over this connection? In other words, none would be the same thing as not using the extension in the first place?

TelegramSam: Yes, but allowing the explicit value allows changing the return-route value on an existing connection. I'll add some clarifying commentary.

dhh1128:
I like this PR and am supportive of merging. I do like Kyle's comment and Sam's intention to add a clarifying sentence or two.

extensions/return_route/main.md
TelegramSam: added. is that clear enough?

kdenhartog (approved):

PR #91, opened 2020-08-28
Revise purpose and scope section

dhh1128:

### DIFF
--- a/scope.md
+++ b/scope.md
@@ -1,10 +1,12 @@
 ## Purpose and Scope
 
-*Initial: Need for a “trust layer for the internet”, and DIDComm’s pivotal role in enabling trustworthy communication based on DIDs. A decentralized version of PKI plus a decentralized version of DNS plus cutting-edge crypto.*
+The purpose of DIDComm is to provide a secure, private communication methodology built atop the decentralized design of [DIDs](https://www.w3.org/TR/did-core/).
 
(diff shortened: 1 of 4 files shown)


scope.md
```diff
@@ -1,10 +1,12 @@
 ## Purpose and Scope
 
-*Initial: Need for a “trust layer for the internet”, and DIDComm’s pivotal role in enabling trustworthy communication based on DIDs. A decentralized version of PKI plus a decentralized version of DNS plus cutting-edge crypto.*
+The purpose of DIDComm is to provide a secure, private communication methodology built atop the decentralized design of [DIDs](https://www.w3.org/TR/did-core/).
```
kdenhartog, line 3: I cringed a bit at the secure, private part because it's only true if VCs are used or a trusted OOB channel is used to setup up. Irregardless, this is definitely an improvement and technically true (with some caveats - but that's what the spec is for.)

kdenhartog (approved):

scope.md
dhh1128: Having a secure pipe is fully achieved by DIDComm. No caveats and no VCs needed. Knowing who's on the other end of that secure pipe is the MITM problem. We need to clarify the difference carefully so expectations are aligned by spec readers. But this is not the place to do something like that.

kdenhartog: Yup, great nuanced point I overlooked. In any case I remain willing to approve.

PR #93, opened 2020-08-31
Revise overview

dhh1128:


scope.md
```diff
@@ -6,24 +6,20 @@ The purpose of DIDComm is to provide a secure communication layer built on top o
...
 
-That's it.
+- An endpoint (web, email, etc) where messages can be delivered to Bob.
+- The public key that Bob's agent is using in the Alice:Bob relationship.
 
-Well, mostly. The description is pretty good, if you squint, but it does not fit all DIDComm interactions:
+Now Alice's agent uses Bob's public key to encrypt the plaintext so that only Bob's agent can read it, adding authentication with its own private key. The agent arranges delivery to Bob. This "arranging" can involve various hops and intermediaries. It can be complex. (See [Routing in the Implementers Guide](/guide#routing).)
+
+Bob's agent eventually receives and decrypts the message, authenticating its origin as Alice using her public key. It prepares its response and routes it back using a reciprocal process (plaintext &#x2192; lookup endpoint and public key for Alice &#x2192; encrypt with authentication &#x2192; arrange delivery).
```
OR13, line 20: This paragraph assumes a default of Alice publishes here service endpoint next to her public key, I wonder if the safer default is that Alice invites Bob to use a service endpoint with her public key....

I am not arguing for or against public service endpoints, but wondering which of the following is the safer default that didcomm should recommend generally:

1. Messages from Alice include a DID URI which resolves to a document with a service endpoint.
2. Messages from Alice include a DID which resolves to a document with a service endpoint.

Put another way, does didcomm have an opinion about the privacy of service endpoints....

Given did comm heritage and focus on privacy, I would expect didcomm to rely on VDR / did resolution for as little as possible, and on authenticated encryption as much as possible.

Consider the following (possibly equivalent / overly technical description)...

Alice visis bob.com, sees that bob buys used car's and that he takes offers via a DID_URL which resolved to a DID Document that contains a service endpoint ( the service endpoint might be in intial-state, it doesn't have to be in the VDR)....

Alice uses Bobs service endpoint to offer to sell her old car, she uses a DID URI which resolves to a did document which includes the service endpoint that BOB should use to reply....

This is 100% compatible with what you have said, and this overview is almost certainly NOT the correct place to say this.... but I wonder if DIDComm should prefer DID_URIs that resolve to documents with service endpoints OVER DIDs that resolve to documents with service endpoints generally.

It seems like the more we can use encryption, and URIs and the less we use VDRs the safer DIDComm , the more privacy aligned it is.

dhh1128: I really like your approach #1 (include DID URI) instead of #2 (include DID that resolves). It is in the spirit of did:peer + initial-state. I probably didn't clue in to the implications of what I wrote because I was assuming peer DIDs, which are likely to be private anyway. But it's a good improvement if we can make the language suggest the more nuanced mental model.

Do you have a revision you'd like to propose? If not, I'll see if I can come up with one that makes the language more aligned with what should be emphasized...

OR13: @dhh1128 I feel this nuance probably longs in the section that first explains what `services` are and how did comm uses them... and not in the intro :) 

I've opened https://github.com/decentralized-identity/didcomm-messaging/issues/94

If you can point out on the issue where such a PR would go, I can take a stab at it.

dhh1128:
Merge conflict resolved.

PR #96, opened 2020-09-10
what's new section

TelegramSam:

### DIFF
--- a/guide/whatsnew.md
+++ b/guide/whatsnew.md
@@ -22,6 +22,12 @@ Each message contains both the sender key (used in the encryption layer), and th
 
 One side effect of the DID Exchange Protocol in V1 was that you confirmed the validity of the DID with a round trip to the other party. Many protocols will provide this assurance via the flow of the protocol prior to the point where round-trip testing is required. When this round-trip is desired prior to the beginning of a protocol, a round trip with another protocol (such as Trust Ping or Feature Discovery) can provide the same assurance.
 
+#### Process From Headers prior to Protocol Processing
+
+Relationship changes in V1 were handled inside the DID Exchange Protocol. In V2, relationship changes including discovery and rotation are handled in message headers.
+
+In V2, messages must evaluate the `from` and `from_prior` headers of every message prior to beginning the protocol message processing. 
+
 #### No technical difference between Ephemeral Mode and Full Mode
 
 Ephemeral mode in V1 was a method of passing messages without first performing an exchange of DIDs. Given that we no longer have a need to perform an exchange of DIDs prior to passing messages of another protocol, we no longer need to designate a mode for ephemeral interactions.
--- a/guide/whatsnew.md
+++ b/guide/whatsnew.md
@@ -1,0 +1,36 @@
+## What's New?
+
+The version of DIDComm incubated in the Hyperledger Aries community is referred to as Version 1 (V1). This spec describes the next version, referred to as Version 2 (V2). This section will describe the changes between V1 and V2, useful to members of the Aries community.
+
+### Summary of Changes
+
+- Formalization of methods used in V1
+  - JWM based envelope
+  - ECDH-1PU standardized form of AuthCrypt
+- Both DID and key in each message
+- Special Handling of Peer DIDs eliminated
+- Message structure split between 'headers' and body.
+- No AnonCrypt encryption method.
+
+### Practical Changes
+
+The list of changes above leads to practical changes in how DIDComm is used.
+
+#### DID Exchange not needed
+
+Each message contains both the sender key (used in the encryption layer), and the sender's DID. The exchange of DIDs that occurs via the DID Exchange Protocol used in V1 occurs in each message that is transferred. Rotating DIDs is accomplished via the `prior_did` header. These features make the DID Exchange Protocol redundant.
+
+One side effect of the DID Exchange Protocol in V1 was that you confirmed the validity of the DID with a round trip to the other party. Many protocols will provide this assurance via the flow of the protocol prior to the point where round-trip testing is required. When this round-trip is desired prior to the beginning of a protocol, a round trip with another protocol (such as Trust Ping or Feature Discovery) can provide the same assurance.
+
+#### No technical difference between Ephemeral Mode and Full Mode
+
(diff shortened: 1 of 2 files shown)


guide/whatsnew.md
```diff
@@ -0,0 +1,36 @@
...
+
+#### No technical difference between Ephemeral Mode and Full Mode
+
+Ephemeral mode in V1 was a method of passing messages without first performing an exchange of DIDs. Given that we no longer have a need to perform an exchange of DIDs prior to passing messages of another protocol, we no longer need to designate a mode for ephemeral interactions.
+
+#### Use Peer DIDs (or other suitable DID method) in place of AnonCrypt
+
+Anoncrypt was a method present in V1 that allowed a message to be encrypted to a recipient without involving the recipient keys. The ease of using Peer DIDs allows the sender to remain anonymous using the existing authenticated encryption method. The encryption properties are the same between the methods. Eliminating this option makes the spec simpler without loosing any features.
```
dhh1128, line 31: I think this description of Anoncrypt is not quite accurate. It allowed a message to be encrypted to the recipient in a way that absolutely *did* involve the recipient keys; what made it interesting is that the sender keys were ephemeral and newly generated for each message (and therefore unrecognizable to the recipient as belonging to anybody in particular).

guide/whatsnew.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Practical Changes
+
+The list of changes above leads to practical changes in how DIDComm is used.
+
+#### DID Exchange not needed
+
+Each message contains both the sender key (used in the encryption layer), and the sender's DID. The exchange of DIDs that occurs via the DID Exchange Protocol used in V1 occurs in each message that is transferred. Rotating DIDs is accomplished via the `prior_did` header. These features make the DID Exchange Protocol redundant.
+
+One side effect of the DID Exchange Protocol in V1 was that you confirmed the validity of the DID with a round trip to the other party. Many protocols will provide this assurance via the flow of the protocol prior to the point where round-trip testing is required. When this round-trip is desired prior to the beginning of a protocol, a round trip with another protocol (such as Trust Ping or Feature Discovery) can provide the same assurance.
```
dhh1128, line 23: I fully agree with this paragraph. It is totally accurate. However, I remain uncomfortable that we're not adequately pointing out the significance of this semantic difference. Under V1, all protocols assumed that a secure pipe already existed and the DID of the first mover in the protocol would be pairwise and stable throughout the protocol run. Under V2, the DID of the initial mover/sender has to change at step 2 of the protocol to achieve the same guarantee, and all protocol impls need to tolerate DID rotation in the middle. If DID rotation is not used, and if the first message is OOB, then external observers can learn the DID and endpoint used by one of the parties. For some ephemeral use cases, this may not be a problem -- but as a general pattern, we have introduced a lurking antipattern that we should be warning against with documentation somewhere.

guide/whatsnew.md
TelegramSam: unfortunate typo: that should read ' involving the sender keys'. I have a reworded fix.

guide/whatsnew.md
TelegramSam: I've cleaned up the language a bit. I think we need a guide section about DID Rotation, and why it must be used under normal conditions to prevent leaking the externally passed DID as you point out.

guide/whatsnew.md
```diff
@@ -0,0 +1,42 @@
...
+
+Relationship changes in V1 were handled inside the DID Exchange Protocol. In V2, relationship changes including discovery and rotation are handled in message headers.
+
+In V2, messages must evaluate the `from` and `from_prior` headers of every message prior to beginning the protocol message processing. 
+
+#### No technical difference between Ephemeral Mode and Full Mode
+
+Ephemeral mode in V1 was a method of passing messages without first performing an exchange of DIDs. Given that we no longer have a need to perform an exchange of DIDs prior to passing messages of another protocol, we no longer need to designate a mode for ephemeral interactions.
+
+#### Use Peer DIDs (or other suitable DID method) in place of AnonCrypt
```
llorllale, line 35: We still cannot use `did:peer` as I said in https://github.com/decentralized-identity/didcomm-messaging/issues/33#issuecomment-680052846.

The `from` and `from_prior` fields would need to be tweaked such that DID URLs be used instead of just DIDs.

And even then, I'm not so sure of the implications of such a tweak...

llorllale: Well... I think the `initial-state` idea was completely scrapped from did-core... so now we depend on `did:peer` to provide the initial state in a DID.

llorllale: See relevant discussions for `did:peer` in:

* https://github.com/decentralized-identity/peer-did-method-spec/issues/24
* https://github.com/decentralized-identity/peer-did-method-spec/pull/23

guide/whatsnew.md
```diff
@@ -0,0 +1,42 @@
...
+
+Anoncrypt was a method present in V1 that allowed a message to be encrypted to a recipient using ephemeral sender keys, allowing the sender to remain anonymous. The ease of using Peer DIDs allows the sender to remain anonymous using the existing authenticated encryption method. The encryption properties are the same between the methods. Eliminating this option makes the spec simpler without loosing any features.
+
+#### Message Level Decorators now represented as Headers
+
+The adjusted structure of DIDComm messages now represents message level decorators as message headers. An example includes `thread_id`.
```
troyronda, line 41: parent thread id will also need to be defined.

guide/whatsnew.md
TelegramSam: given wg call discussion on 2020/09/14, the DIDComm JWM profile needs to be adjusted to accommodate whatever method the DID spec/community chooses to pass off/pre ledger recorded information.

guide/whatsnew.md
TelegramSam: given wg call discussion on 2020/09/14, we will update the JWM DIDComm Profile to pull forward thread_id as well as adding parent threads.

guide/whatsnew.md
kdenhartog: As we talked about on the WG call, when we change to a DID URL we'll want to specify that during the "from" validation process the DID must be resolved to a full did document rather than de-referenced to only the key included in that DID URL. I agree that we're going to need to carry the parameters through though and changing from DID to DID-URL would allow this to occur.

tplooker (approved):

PR #97, opened 2020-09-10
trust ping protocol

TelegramSam:

### DIFF
--- a/trustping.md
+++ b/trustping.md
@@ -1,0 +1,78 @@
+# Trust Ping Protocol 1.0
+
+A standard way for agents to test connectivity,
+responsiveness, and security of a DIDComm channel.
+
+## Motivation
+
+Agents are distributed. They are not guaranteed to be
+connected or running all the time. They support a
+variety of transports, speak a variety of protocols,
+and run software from many different vendors.
+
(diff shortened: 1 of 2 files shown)


trustping.md
```diff
@@ -0,0 +1,78 @@
...
+{
+  "type": "https://didcomm.org/trust_ping/1.0/ping",
+  "id": "518be002-de8e-456e-b3d5-8fe472477a86",
+  "body": {
+      "response_requested": true
+  }
+}
+```
+
+**response_requested**: default value is `true`. If false, the `sender` is not requesting a `ping_response` from the `receiver`. If `true`, the `sender` is requesting a response.
```
llorllale, line 55: > If false, the `sender` is not requesting a `ping_response` from the `receiver`.

Can you provide an example use case?

TelegramSam: This has been used in practice in v1 as a way to send an otherwise benign signed message. For example, an opened websocket transport can send a one-way ping, and the receiving agent can validate the key used to the connection, and forward queued messages

llorllale:
@TelegramSam I think I recall us reaching an agreement on a call a while back to accept this one as is?

Are there any blockers?

PR #98, opened 2020-09-10
feature discovery

TelegramSam:


feature_discovery.md
```diff
@@ -0,0 +1,158 @@
...
+
+##### `query` Message Type
+
+A `discover-features/query` message looks like this:
+
+```json
+{
+  "@type": "https://didcomm.org/discover-features/1.0/query",
+  "@id": "yWd8wfYzhmuXX3hmLNaV5bVbAjbWaU",
+    "body": {
```
llorllale, line 48: nit: indentation is off

feature_discovery.md
```diff
@@ -0,0 +1,158 @@
...
+it imputes to the other party based on cumulative interactions so far,
+whether it's in the middle of upgrading a plugin, whether it's currently
+under high load, and so forth. And responses to a `discover-features` request are
+not guaranteed to be true forever; agents can be upgraded or downgraded,
+although they probably won't churn in their protocol support from moment
+to moment.
+
+### Privacy Considerations
+
+Because the regex in a `request` message can be very inclusive, the `discover-features`
```
llorllale, line 138: > regex

* we need to be very clear about the format of the contents of `query`. The concept of "regular expressions" appears once in this text, only here, and not in the section that defines `query`
* having said the previous - regular expressions seem like overkill... I forsee most implementations using just wildcards  on protocol names and versions... not sure about the authority... are extensions allowed under `didcomm.org`?

feature_discovery.md
TelegramSam: fixed.

feature_discovery.md
kdenhartog: I'd prefer that we not suggest regex as a method to process the different features. It opens up the possibility for [ReDoS attacks](https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS) for implementers of this feature.

feature_discovery.md
```diff
@@ -0,0 +1,158 @@
...
+use `discover-features` to answer legitimate questions, and not to build detailed
+profiles of one another. However, fingerprinting may be attempted
+anyway.
+
+For agents that want to maintain privacy, several best practices are
+recommended:
+
+##### Follow selective disclosure.
+
+Only reveal supported features based on trust in the relationship.
```
kdenhartog, line 151: How do we codify and test for "trust in the relationship"? Having more details here could be useful as to the expected way that an agent would determine whether it should send a subset of features on it's list or the whole list.

feature_discovery.md
TelegramSam: It isn't actually a regex, just wildcards (*)

feature_discovery.md
TelegramSam: Why certain feature should be revealed is quite specific to the implementation and the features themselves. I do understand the desire to end up with testable statements. Should discussion of selective disclosure belong in the guide instead of the spec?

feature_discovery.md
kdenhartog: Thanks for clarifying. That makes sense. 👍

feature_discovery.md
kdenhartog: Agreed on call this week that this would be good to move to the implementation guide. I'm fine with leaving this statement in the spec if you want as well.

tplooker (approved):

tplooker:
Ready to merge as per WG consensus, however conflicts need to be resolved

PR #117, opened 2020-09-18
add details about perfect forward secrecy usage

kdenhartog:


awoie (approved):
LGTM

TelegramSam:
This is an awesome addition. Given our discussion on killing Anoncrypt (ECDH-ES), should we include that part of the explanation?

encryption.md
```diff
@@ -58,8 +58,14 @@ TODO: Include language about safe nonce considerations.
...
 
-### Examples
+## Perfect Forward Secrecy
+
+Due to the triple Key Derivation algorithm used in ECDH-1PU and the ephemeral keys used per message in ECDH-ES, all messages sent via DIDComm have perfect forward secrecy without any additional security added by the transport layer. In ECDH-ES, PFS is obtained because the ephemeral key combined with the static recipient key produce a new derived symmetrical key for each message (which is then used to encrypt the content encryption key).
+
+In the case of ECDH-1PU this is achieved by encrypting the content encryption key with the output of the hash of the Ze (ECDH of Ephemeral key and recipient static key) and Zs (ECDH of static sender key and recipient static key). With Ze bringing the changed derived secret in each message and Zs bringing the repudiable authenticity of each message, the resulting Z (hash of Ze and Zs) carries the properties of perfect forward secrecy and repudiable authenticity for each message as well.
```
TelegramSam, line 65: Further reading of this has me questioning: if static sender keys and static recipient keys are used, then doesn't that not provide PFS? If those static keys are compromised, then other messages could still be decrypted.

encryption.md
```diff
@@ -58,8 +58,14 @@ TODO: Include language about safe nonce considerations.
 | ECDH-1PU+A256KW | P-521           | EC                 | ECDH-1PU key wrapping using key with NIST defined P-521 elliptic curve to create a 256 bits key as defined in [ecdh-1pu](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03#section-2.2)                                                         |
 | ECDH-1PU+A256KW | X25519          | OKP                | ECDH-1PU X25519 ([RFC7748 section 5](https://tools.ietf.org/html/rfc7748#section-5)) to create a 256 bits key as defined in [ecdh-1pu](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03#section-2.2)                                           |
 
-### Examples
+## Perfect Forward Secrecy
+
+Due to the triple Key Derivation algorithm used in ECDH-1PU and the ephemeral keys used per message in ECDH-ES, all messages sent via DIDComm have perfect forward secrecy without any additional security added by the transport layer. In ECDH-ES, PFS is obtained because the ephemeral key combined with the static recipient key produce a new derived symmetrical key for each message (which is then used to encrypt the content encryption key).
```
mikelodder7, line 63: This is only true for the sender. The recipient does not get PFS.

mikelodder7: This is called weak forward secrecy not perfect forward secrecy. We get perfect if both sides use ephemeral keys.

kdenhartog: Thanks for the feedback. I'll be sure to update to describe this better.

kdenhartog:
Alright @TelegramSam this should now be properly updated. Thanks @mikelodder7 for reviewing and making sure we're describing things correctly.

encryption.md
kdenhartog: Since the message is encrypted with the symmetrical key created via a combination of ephemeral keys and static keys on the sender side, they do get PFS. However, since the recipient uses a static key they're left with no PFS. Based on Mike's feedback I understood this to be referred to as weak perfect forward secrecy in traditional academic literature.

PR #119, opened 2020-09-25
slack link update

dif-admin:


TelegramSam (approved):
thanks for the clarification.

PR #122, opened 2020-10-12
Future Encodings Language

TelegramSam:


TelegramSam:
DIDComm Messages are JSON encoded and relies on the Jose family of specs at the encryption, signature, and content level. Future encodings MUST specify a deterministic and reliable method for indicating the alternative encoding used.

kdenhartog (approved):

Issue #118, opened 2020-09-21
need guidance about structured headers


dhh1128:
With the DIDComm v1-->v2 shift in emphasis from decorators to headers, we seem to be thinking about headers as individual scalar items (a flat array of headers that are strings or numbers) in much the same way that HTTP headers work. For example, we appear to be headed toward separate top-level headers for `pthid` and `id` rather than a `thread` header that contains `pthid` and `id`.

I feel uneasy about this as a loose mental drift that hasn't been analyzed, but I could be okay with it as an official policy that we've explicitly aligned around. Or not. Do we want a flat header space, or do we want structured headers? Why or why not?

A virtue of flat headers is that they're really easy to model as a data structure (a map of string -> StringOrInt). A drawback is that they don't allow arrays or complex values; you have to do secondary parsing on strings that contain substructure. We already have at least one array value in our currently defined headers... HTTP headers have an uneasy relationship with this issue; consider the funky syntax of strings with the `Accept` header in content negotiation.

A virtue of structured headers is that they can use ordinary JSON parsing to give you structure (correcting the problem with flat scalar values mentioned above, without any extra burden on parsing logic). Also, they can express relatedness and co-occurrence rules, so you don't end up with combinations of headers that make no sense.

Here is a first cut at some guidance about structured headers that we might include in the spec. What input do you have about it?

----------
Headers can be simple (mapping a header name to an integer or a string) or structured (mapping a header name to JSON substructure -- an array or JSON object). When defining a new header type, the following rules guide how its value is modeled:

* Headers SHOULD NOT use more structure than necessary; simple headers are preferred.
* However, a header value SHOULD NOT require interpretation over and above ordinary JSON parsing (e.g., specialized string syntax) before it is usable.
* Also, headers that are only meaningful together SHOULD be grouped into a JSON object.

[TODO: give examples that explain why it's okay to have base64url-encoded binary, IP addresses, email addresses, etc -- but not some other stuff.]

dhh1128:
tagging @TelegramSam , @kdenhartog , @llorllale , @swcurran , and @OR13 , who might have opinions.

OR13:
we had a similar discussion regarding did document metadata / resolution input options....

If the headers don't support full JSON, you will see this __everywhere__

```
const vendorHeader = base64url.encode(JSON.canonicalize({ ...bigJsonObject }}));
```

[...]

PR #127, opened 2020-11-10
Add guidance about structured headers.

dhh1128:
Addresses the community consensus that emerged on issue #118.


kdenhartog (approved):
Looks good to me. I was a bit hesitant upon first reading it (only because I wasn't sure if we wanted to make this an extension point in DIDComm) but convinced myself this is the correct direction to go to allow flexibility in the routing extensions that will be built on top.

TelegramSam:
Daniel will add a minor clarifying update.

Issue #114, opened 2020-09-17
Sec 3 - to attribute defined as optional


downTheFallLine:
Current specification indicates that to attribute is optional.  If a message is created without a "to", what does that mean?  Should this attribute be required?

TelegramSam:
If the message is signed, but not encrypted, the 'to' may not be known. We could clarify that this MUST be specified in an encrypted message.

downTheFallLine:
I'm still confused.   If a message has no "to" attribute, where does it go?  How would it be routed?  To whom would it be routed?

TelegramSam:
The 'to' attribute is not useful in message that is signed, but not encrypted. This is useful when you want to pass a message to an unknown recipient. A current example is the [Out of Band Invitation](https://identity.foundation/didcomm-messaging/spec/#out-of-band-messages) message which is passed via link or QR code. Other methods are also possible.

Messages without a 'to' attribute are NOT capable of being routed.

kdenhartog:
The `to` value isn't needed for the final recipient of any message irregardless of the envelope format as well. 'to' remains in there for forward messages so that mediators are able to properly route payloads.

TelegramSam:
(WG 20201109.) forward message 'to' attrribute needs a rename. perhaps destination/next?

dhh1128:
I believe this issue can be closed once PR #128 is merged.

PR #128, opened 2020-11-10
Clarify semantics of `to` header

dhh1128:
Responds to the conversation around issue #114


message_structure.md
```diff
@@ -15,18 +15,23 @@ The following example shows common elements of a DIDComm Message. Further detail
...
-- **to** - OPTIONAL. Recipient(s) identifier. The `to` attribute MUST be an array of strings where each element is a valid [DID](https://w3c.github.io/did-core/#generic-did-syntax) which identifies the recipients of the message.
+- **to** - OPTIONAL. Identifier(s) for recipients. MUST be an array of strings where each element is a valid [DID](https://w3c.github.io/did-core/#generic-did-syntax) that identifies a member of the message's intended audience.
+
+  The relationship between this array and routing is worthy of comment. A message from Alice to Bob is seen as plaintext by Alice and by Bob, and as ciphertext by all intermediaries. Thus, the value of the `to` array as seen by Bob cannot be used for routing. Instead, when Bob sees it, it offers semantics like the `To:`, `CC:`, and `BCC:` headers in email, allowing a recipient to enumerate all parties that the sender claims might see the message in plaintext. This has semantic value in application-level protocols built atop DIDComm.
```
kdenhartog, line 31: I was just thinking about this and realized that ever mediator/relay is going to need to have a mapping of the publicKey values in the recipient block of the JWE/JWS in order to know which endpoint to send the message to. However, if we model this `to` value in a way that does allow it to be used in routing then the mediator/relay can dynamically rely on the ability to resolve a did document to get a service endpoint for the recipients rather than keeping a pre-registered mapping for everyone they're capable of routing messages for. Is that something that we find is valuable? If so I would think we want to reword this a bit.

dhh1128: I'm not sure whether I am understanding how you view routing, Kyle. I've built a diagram that might help us test alignment. Note how I'm imagining the `to` attribute will be used.
Screen Shot 2020-11-12 at 5 32 01 PM

In case you want to create a different version of the same thing, here's the original: https://docs.google.com/presentation/d/1L-lbj6aHrbtprA1fa4BsJBxNtW3YS_D811auuxfoevY/edit

TelegramSam: @dhh1128 I only understand the diagram above if the message displayed are plaintext messages, and encrypted forms aree not represented.

I think the right answer here is for the `forward` message to have a `to` attribute which contains a DID. The mediator will use this for routing, and doesn't need to maintain a key mapping. This does make it 'redundant' with the `to` attribute of the didcomm plaintext message - perhaps we should rename the forward message's to attribute to avoid redundancy.

kdenhartog: Sorry, just realized I responded with only a thumbs up rather than sending the response I had written up. With the usage of your graphic I believe we're still on the same page @dhh1128 . Originally, I was thinking you were proposing to remove the routing form of the `to` in favor of the semantic form of `to`, but from the looks of it you're actually just wanting to add a new property to express the semantics of the who the recipient is intended for in the final message. Am I reading that correctly?

> @dhh1128 I only understand the diagram above if the message displayed are plaintext messages, and encrypted forms aree not represented.

which message are you thinking isn't represented when encrypted? I don't think I quite grokked what you were pointing out here.

> I think the right answer here is for the `forward` message to have a `to` attribute which contains a DID. The mediator will use this for routing, and doesn't need to maintain a key mapping. This does make it 'redundant' with the `to` attribute of the didcomm plaintext message - perhaps we should rename the forward message's to attribute to avoid redundancy.

I was thinking that the `to` header would be used in a forward style (m[1], m[2], m[3]) message, but in the plaintext message (inside m[0] I think in this graphic) we would drop it. However, based on my reading @dhh1128 would like to have a "semantic to" so that the final recipient knows that the message was for them which is why I'm understanding that we want to define them as separate properties because in one instance the `to` is used for routing and in the other it's used for semantics to describe the recipient. Is that where you're going with this @TelegramSam ?

dhh1128: I don't think we can resolve this in a written comment stream. Can we discuss on the next DIDComm WG meeting, or maybe in a special call?

message_structure.md
```diff
@@ -15,18 +15,23 @@ The following example shows common elements of a DIDComm Message. Further detail
...
+  
+  However, when the innermost message from Alice to Bob is attached as encrypted payload to a `forward` message in DIDComm's routing protocol, a mediator sees the `to` header of the `forward` message in plaintext. Thus, the `to` header in one or more containing envelopes does play a role in routing.
```
TelegramSam, line 33: This is incorrect, and needs to be removed.

message_structure.md
```diff
@@ -15,18 +15,23 @@ The following example shows common elements of a DIDComm Message. Further detail
...
 ### Message Headers
 
-The attributes of a DIDComm Message at the level of its outer packaging (effectively, the "headers" of the message) are as follows:
+The attributes of a DIDComm plaintext message at the level of its outermost object (effectively, its "headers") are as follows:
 
 - **id** - REQUIRED. Message ID. The `id` attribute value MUST be unique to the sender.
 - **type** - REQUIRED. Message Type. The `type` attribute value MUST be a valid [Message Type URI](protocols.md#message-type-uri) , that when resolved gives human readable information about the message. The attributes value also informs the content of the message, for example the presence of other attributes and how they should be processed.
-- **to** - OPTIONAL. Recipient(s) identifier. The `to` attribute MUST be an array of strings where each element is a valid [DID](https://w3c.github.io/did-core/#generic-did-syntax) which identifies the recipients of the message.
+- **to** - OPTIONAL. Identifier(s) for recipients. MUST be an array of strings where each element is a valid [DID](https://w3c.github.io/did-core/#generic-did-syntax) that identifies a member of the message's intended audience.
+
```
TelegramSam, line 30: indicate what is implied if the to attribute is missing.

dhh1128: Done.

message_structure.md
dhh1128: I largely removed this, but kept one sentence explaining why using `to` with routing doesn't make sense. I think it's simplified but clear.

kdenhartog (approved):

TelegramSam (approved):

tplooker (approved):

PR #129, opened 2020-11-13
Add info about post quantum

dhh1128:


future_proofing.md
```diff
@@ -1,21 +1,33 @@
...
+
+### Beyond Messaging
+
+This is a DIDComm _messaging_ spec. Security, privacy, routing, and metadata concepts from this spec could be adapted to other communication styles, including multicast/broadcast and streaming. This will create sister specs to DIDComm Messaging, rather than evolving DIDComm Messaging itself. 
+
+### Post-Quantum Crypto
+
+The designers of DIDComm are aware that DIDComm's cryptographic methods will need to be upgraded when quantum computing matures. This is because DIDComm makes heavy use of asymmetric elliptic curve mechanisms that depend on the discrete logarithm problem; this computational hardness is likely to be vulnerable to quantum cracking. Similar risks will drive upgrades to TLS, Ethereum, Bitcoin, and many other systems that are considered highly secure today.
+
+Some foundations for quantum-resistant DIDComm have already been laid. DIDComm is able to use arbitrary DID methods. KERI's prerotation scheme (implemented in did:un) is quantum-secure, and peer DIDs support arbitrary algorithms including ones that are quantum-secure.
```
kdenhartog, line 29: KERI would not remain quantum secure in the context of DIDComm if I understand the properties it gives rise to properly. From my understanding, because KERI relies upon a hash (which have no known post-quantum attacks) the commitment scheme remains quantum secure. However, this protection is only valid for the quantum security which is orthogonal to the usage of the keys in DIDComm.

I'd prefer that we not allude to DIDComm being PQS at this point and remain silent on the topic because everything within our realm has been focused on using classical computing cryptographic algorithms to date and no one has worked to integrate the PQ digital signature algorithms into the JOSE family yet.

future_proofing.md
```diff
@@ -1,21 +1,33 @@
...
-### Encoding
+### Future Work
 
-DIDComm Messages are JSON encoded (based on the JOSE family of specs) at the encryption, signature, and content level. Future encodings MUST specify a deterministic and reliable method for indicating the alternative encoding used.
\ No newline at end of file
+#### Additional Encodings
+
+DIDComm Messages are JSON encoded (based on the JOSE family of specs) at the encryption, signature, and content level. Future encodings might introduce binary serializations. Each innovation like this MUST specify a deterministic and reliable method for indicating the alternative encoding used.
+
+### Beyond Messaging
```
kdenhartog, line 21: This is a good addition in my opinion and I think it would be wise to keep it in here no matter the outcome of the discussion on Post-Quantum Crypto.

future_proofing.md
dhh1128: @SmithSamuelM: do you have an opinion about the reference to KERI here?

dhh1128: @kdenhartog : I updated the verbiage slightly. Do you like it better in the new form?

kdenhartog: Yup happy with how you re-worded it now and believe it would be useful for us to have in here now.

kdenhartog (approved):
After the changes made, I'm good with how this PR reads.

future_proofing.md
SmithSamuelM: KERI's post quantum security to recover compromised keys with un-compromisable digest commitments to pre-rotated keys. So an attack on DIDcom could succeed in compromising the current set of signing keys which might have some adverse effects but the controller could recover and resume with new keys. So its an open question of what that means from a DIDCom harm perspective.

kdenhartog: Ok, I think that it does remain a bit orthogonal then because if an attacker was able to compromise the signing keys with a Quantum Computer then it would likely invalidate the guarantees that the digital signatures are intending for us around who the author of the message is (just as any other style Key compromisation attack would). The only difference as you point out is that KERI is setup to properly recover and keep the same identifier for later usage where as others would likely have to rotate their DID.

In this case, I think it's useful to not make mention of KERI here and leave the text as is (with the new changes that are beneficial to make mention of).

Issue #134, opened 2020-12-14
`reply_*` headers and the support of DIDs without endpoints


TelegramSam:
Currently, the spec requires a DID used to have an endpoint. Any DID without an endpoint would need to rotate to one that supports endpoints (including did:peer) to engage in DIDComm communication.

JWM message support `reply_to` and `reply_url` that could be used to support dynamic supply of endpoint information. Adding this does have a carrying cost of state management and message sending complexity.

What should we do?

TelegramSam:
Simplest proposal: leave as is. Add guide section describing how did rotation can be used to provide a provenance chain from an existing endpointless DID.

This leaves adding this as work for a potential extension or feature for next version if there is sufficient demand.

chunningham:
For methods which support ephemeral or non-global Identifiers without support for service endpoints (basically `did:key` and KERI in direct mode) I would say that `reply_*` would make it much easier to integrate DIDComm without adding the extra step of a `did:peer` masquerade (even if both of those examples do map fairly well to `did:peer`).

Additionally, is it possible that there may be use cases where a party may want to change endpoint used for DIDComm? In that case a `reply_*` would be much simpler (imo) than rotating to a new DID with the new endpoint or performing a `did:peer` update which adds/changes an endpoint.

TelegramSam:
Also discussed in the meeting: There are a number of edge cases with changing service endpoints that are very powerful, but can be sufficiently solved with a protocol. Solving a problem with headers should only be done when a protocol is infeasible.

Example: If Alice and Bob have exchanged DIDs and know each other's endpoints, then end up in close proximity to each other temporarily, a protocol message can facilitate use a bluetooth connection for a limited time.

awoie:
@TelegramSam I guess `reply_*` would also not help in the following situation where I have a vanilla `did:ethr` (created off-chain) that is essentially just a secp256k1 public key which does not have a service endpoint. _Note: for `did:ethr` to have any secure encryption key such as X25519 keys and a service endpoint, a ledger transaction would be necessary._ Because I don't have a valid encryption key, I would need to rotate to another DID anyways even if `reply_*` was available. The DID rotation could occur in the very first DIDComm message. If one does not want to rotate away from `did:ethr`, they would need to add a secure encryption key and a service endpoint. Would you agree with that?

llorllale:
> If one does not want to rotate away from did:ethr, they would need to add a secure encryption key and a service endpoint. Would you agree with that?

[...]

PR #139, opened 2020-12-29
Proposed reply header text

TelegramSam:
In our WG meeting, we decided to propose text for reply header handling. This is that proposal.

I am not advocating specifically for this: it is my intent to explore what is required to allow reply processing, including evaluating the carrying costs of this approach. Everything in this PR is subject to change/improvement/refinement. 

Relates to #134


TelegramSam:
In WG Call 01042021 we discussed a possible alternative: What if the endpoint was expressed as an 'endpoint_override' instead of as the 'reply' method included in this PR? Could that eliminate some of the shortcomings?

llorllale:
This path deviates away from the approach advocated for in DIDComm v1 a couple of weeks back which was to use a single identifier to convey the necessary parameters for the recipient to reply. It was argued back then that such an approach would essentially build a bridge into V2.

That conversation took place a week or two before this PR was pushed. Is this PR still the current thinking or have we switched away from this and on to the direction I refer to above?

TelegramSam:
I propose we kill this PR. Let's discuss in WG meeting today.

TelegramSam:
Killed per conversation in WG 20210614

PR #140, opened 2020-12-29
Updated missing expires_time semantic

### LINKED ISSUE #116: Sec 3 - "expires_time" is optional
The spec indicates that "expires_time" is optional.   In the event a message is constructed without this attribute, is there an implied semantic that the message has no expiration.  If so, the implied semantic should be explicitly stated, in my view.

TelegramSam:
Added the semantics, as well as a warning on the top of the jwm file.
Fixes #116 


message_structure.md
```diff
@@ -89,7 +89,7 @@ The predefined attributes of a DIDComm plaintext message at the level of its out
...
 
-- **expires_time** - OPTIONAL. Message Expired Time. The `expires_time` attribute is used for the sender to express when they consider the message to be expired, expressed in UTC Epoch Seconds (seconds since 1970-01-01T00:00:00Z UTC) [link](1970-01-01T00:00:00Z UTC). This attribute signals when the message is no longer valid, and is to be used by the recipient to discard expired messages on receipt.
+- **expires_time** - OPTIONAL. Message Expired Time. The `expires_time` attribute is used for the sender to express when they consider the message to be expired, expressed in UTC Epoch Seconds (seconds since 1970-01-01T00:00:00Z UTC) [link](1970-01-01T00:00:00Z UTC). This attribute signals when the message is no longer valid, and is to be used by the recipient to discard expired messages on receipt. When omitted, the message is considered to have no expiration by the sender.
```
dhh1128, line 92: Question: Suppose a message is marked to expire at noon. Agent X receives the message 5 seconds before noon, and then begins a process that will take 30 seconds before it can be fully processed. Should Agent X discard the message at noon, or continue to process it because it received the message before noon?

The way I defined this in the original timing RFC was that expires time is the moment at which the sender begins to repudiate the message (acts as if it was never sent) -- NOT the time at which the message should be discarded by the recipient. The reason I went with this is that I think the sender can't predict and isn't interested in how long it might take the recipient to reply/process; they can only say when it's too late from their perspective. An agent that receives a message at 11 am with expiration_time of noon should discard the message immediately, if they know it will take 2 hours to process.

dhh1128 (commented):
Not sure about the semantics implied by the description. The specific wording is fine, and the intent is fine, but see my inline question.

awoie:
As per DIDComm WG meeting (Jan 25), merging this PR.

PR #141, opened 2020-12-29
Anonymous From clarification

### LINKED ISSUE #115: Sec 3 - from "attribute" is optional
The spec states "from" as an optional attribute.  If the attribute is missing from a message, and the message is to be encrypted and/or signed, how then does the recipient go about determining key resolution?  Is this handled elsewhere in the spec?  If so, could we at least add a reference in this section?

TelegramSam:
Explains how to remain anonymous while specifying 'from'.
Fixes #115


dhh1128 (approved):

tplooker (approved):

tplooker:

PR #150, opened 2021-01-04
Adds language about VDR disclosure.

### LINKED ISSUE #94: DIDComm should recommend against unnecessary disclosure of services to the VDR
While its possible to resolve both a DID and a DID URL to a DID Document with a service suitable for didcomm...

We should prefer not to disclose services when possible to the VDR...

Reasons for not including services in VDR are documented here: https://github.com/w3c/did-core/issues/382

DIDComm should not prohibit the use of DIDs that publicly publish service endpoints (personal opinion, neither should DID Core....)., however we can recommend against ANY unnecessary disclosure outside of an encrypted channel... 

We can easily support both, and warn users about how if they publish their personal web server to bitcoin... its gonna be there forever.... with no ability to delete.... GDPR, considerations apply...

TelegramSam:
Fixes #94

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -48,7 +48,7 @@
       "spec_directory": "./guide",
       "markdown_paths": [
         "intro.md",
-		"privacy.md",
+	"privacy.md",
         "routing.md",
         "whatsnew.md"
       ]
--- a/specs.json
+++ b/specs.json
@@ -48,6 +48,7 @@
       "spec_directory": "./guide",
       "markdown_paths": [
         "intro.md",
+		"privacy.md",
         "routing.md",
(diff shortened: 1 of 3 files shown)


dhh1128 (approved):

specs.json
```diff
@@ -48,6 +48,7 @@
       "spec_directory": "./guide",
       "markdown_paths": [
         "intro.md",
+		"privacy.md",
```
tplooker, line 51: Need to fix this indent?

tplooker, line 51:
```suggestion
	"privacy.md",
```

PR #151, opened 2021-01-04
Clean up routing forward message examples

### LINKED ISSUE #130: fix forward message structure
add an attribute for 'next' destination.
update to v2 structure

TelegramSam:
Also removes advanced routing options.
Fixes #130

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -38,65 +38,32 @@ The only message in this protocol is the `forward` message. A simple and common 
 
 ```json
 {
-    "@type": "https://didcomm.org/routing/2.0/forward",
-    "to"   : "did:foo:1234abcd",
-    "payloads~attach": [
-        // One payload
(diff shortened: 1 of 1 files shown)


dhh1128 (approved):
I think these simplifications are good for now. However, I would like to revisit them a little bit when we create a section on timing. Expires_time is clearly the most important of these properties; however, more sophisticated properties like wait_until_time or delay_milli also feel interesting to me. We may not need them now -- but I would hate to make expires_time an independent, root-level header, only to discover down the line that we want to introduce some additional timings and therefore want to group them together. Maybe it would be better to have timing { expires_time: value} in these examples, omitting the fancier stuff but creating a place for them to go?

I am not sure about that recommendation, though, so I don't think this should hold up the merge. Let's get this improvement in and then, if we have time and inclination, polish it later to the extent that it matters. It may not matter at all.

tplooker:

PR #153, opened 2021-01-09
Remove anoncrypt from the encryption section.

### LINKED ISSUE #92: What if V2 didn't have AnonCrypt?
In DIDComm V1, we have Authcrypt and Anoncrypt. Anoncrypt follows much the same form as Authcrypt, but instead of a known sender key, an ephemeral key is used. With peer DIDs being easy to create, an Authcrypt message using a peer did is basically equivalent to an Anoncrypt message.

If they are equivalent, why don't we eliminate Anoncrypt as an option in V2, for the benefit of simplicity?

TelegramSam:
Fixes #92

### DIFF
--- a/encryption.md
+++ b/encryption.md
@@ -1,33 +1,8 @@
 # Message Encryption
 
-DIDComm supports two types of message encryption: Authenticated Sender Encryption and Anonymous Sender Encryption. Both forms are encrypted to the recipient, but only Authenticated Sender Encryption provides assurances of who the sender is.
-
 The encrypted form of a JWM is a JWE. The JOSE family defines [JSON Web Algorithms](https://tools.ietf.org/html/rfc7518) (JWAs) which standardize certain cryptographic operations that are related to preparing JOSE structures. For the purposes of interoperability, DIDComm messaging does not support all JWAs, rather it takes a subset of the supported algorithms that are applicable for the following cases around secure messaging. These supported algorithms are listed here.
 
-
-## Sender Authenticated Encryption
-
-When a sender would like to encrypt a message for a recipient or multiple recipients and also be authenticated such that the recipients can determine they were the party that encrypted the message, the JWA of `ECDH-1PU` defined by [draft](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03) SHOULD be used within the structure of a JWE.
-
-For the keys involved in key agreement, the following elliptic curves MUST be supported.
-
-| Curve  | Description                                                  |
-| ------ | ------------------------------------------------------------ |
-| X25519 | The underlying curve is actually `Curve25519`, however when used in the context of Diffie-Hellman the identifier of `X25519` is used |
-| P-256  | NIST defined P-256 elliptic curve                            |
-
-For content encryption of the message, the following algorithms MUST be supported.
-
-| Algorithm(JWA) | Description                |
-| -------------- | -------------------------- |
-| XC20P          | XChaCha20Poly1305          |
-| A256GCM        | AES-GCM with a 256 bit key |
-
-TODO: Include language about safe nonce considerations.
-
-## Anonymous Encryption
-
-When a sender would like to encrypt a message for a recipient or multiple recipients but not be authenticated by the recipients as the party who encrypted the message, the JWA of `ECDH-ES` defined by [RFC 7518](https://tools.ietf.org/html/rfc7518#section-4.6) SHOULD be used within the structure of a JWE.
+For an encrypted DIDComm Message, the JWA of `ECDH-1PU` defined by [draft](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03) SHOULD be used within the structure of a JWE.
 
 For the keys involved in key agreement, the following elliptic curves MUST be supported.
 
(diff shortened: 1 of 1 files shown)


encryption.md
```diff
@@ -25,38 +20,18 @@ For content encryption of the message, the following algorithms MUST be supporte
...
-| ECDH-1PU+A256KW | P-384           | EC                 | ECDH-1PU key wrapping using key with NIST defined P-384 elliptic curve to create a 256 bits key as defined in [ecdh-1pu](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03#section-2.2)                                                         |
-| ECDH-1PU+A256KW | P-521           | EC                 | ECDH-1PU key wrapping using key with NIST defined P-521 elliptic curve to create a 256 bits key as defined in [ecdh-1pu](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03#section-2.2)                                                         |
-| ECDH-1PU+A256KW | X25519          | OKP                | ECDH-1PU X25519 ([RFC7748 section 5](https://tools.ietf.org/html/rfc7748#section-5)) to create a 256 bits key as defined in [ecdh-1pu](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03#section-2.2)                                           |
+When the sender wishes to remain anonymous, they should use an unused DID for the interaction. When the sender is done with the interaction, they can abandon further use of that DID.
```
dhh1128, line 34: The phrase "they should use an unused DID" feels like it might make people's foreheads wrinkle, although I know what you mean.

Maybe reword to "they should identify themselves as a sender using a newly minted DID that's never been seen before. When the sender..."?

dhh1128 (approved):
I think this is a great change. I have one wording suggestion which you may or may not want to apply -- I vote to merge it either way.

encryption.md
Baha-sk: interesting, we're using a new DID to represent a anonymous (technically a fresh new user) user

Baha-sk:
There's one caveat to using ECDH-1PU (and I've exchanged a couple e-mails with the draft author about it), the sender public key is not to be shared inside the JWE envelope. This means the recipient should have the sender key (and therefore probably their DID as well) prior to decrypting the envelope. The sender identity must not be revealed during the message transmission of the envelope. ie the recipient should be able to resolve the sender's key "_out of band_".

This is why the new `skid` recipient header was introduced.

kdenhartog:
> There's one caveat to using ECDH-1PU (and I've exchanged a couple e-mails with the draft author about it), the sender public key is not to be shared inside the JWE envelope. This means the recipient should have the sender key (and therefore probably their DID as well) prior to decrypting the envelope. The sender identity must not be revealed during the message transmission of the envelope. ie the recipient should be able to resolve the sender's key "_out of band_".
> 
> This is why the new `skid` recipient header was introduced.

That's true if we were attempting to keep authentication, but in this case we're not because it's the anoncrypt version. Effectively what we're looking to achieve with this change is to drop the need for an ECDH-ES implementation, but keep the same functionality by using an ephemeral DID rather than a DID known by the recipient.

Also, I've been meaning to follow up with the author about this to check that the skid is not actually a commonly sent ID, but rather an encrypted version of it that is encrypted with the sender's ephemeral key and the recipients static key. This way the skid remains anonymous to outside participants monitoring the communication patterns (keeping in line with our original intent of authcrypt).

encryption.md
```diff
@@ -1,13 +1,8 @@
...
-
 The encrypted form of a JWM is a JWE. The JOSE family defines [JSON Web Algorithms](https://tools.ietf.org/html/rfc7518) (JWAs) which standardize certain cryptographic operations that are related to preparing JOSE structures. For the purposes of interoperability, DIDComm messaging does not support all JWAs, rather it takes a subset of the supported algorithms that are applicable for the following cases around secure messaging. These supported algorithms are listed here.
 
-
-## Sender Authenticated Encryption
-
-When a sender would like to encrypt a message for a recipient or multiple recipients and also be authenticated such that the recipients can determine they were the party that encrypted the message, the JWA of `ECDH-1PU` defined by [draft](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03) SHOULD be used within the structure of a JWE.
+For an encrypted DIDComm Message, the JWA of `ECDH-1PU` defined by [draft](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03) SHOULD be used within the structure of a JWE.
```
kdenhartog, line 5: Could we upgrade this to a MUST rather than a SHOULD now that ECDH-1PU is the only supported JWA? I believe originally this was only a SHOULD rather than a MUST because we supported the use of ECDH-ES, but with that gone, I think it would be more useful for us to upgrade this to a MUST.

As a side note, we're going to need to make sure this draft makes it into a WG as well. Would love for others to help assist with that effort if anyone has contacts at IETF that may be able to help with that.

encryption.md
TelegramSam: Love the wording suggestion.

encryption.md
TelegramSam: changed SHOULD to MUST. Great point.

tplooker:

PR #157, opened 2021-02-02
added context definitions

awoie:
Example usage in DID Docs:

```
"@context":[
    "https://www.w3.org/ns/did/v1",
    "https://identity.foundation/didcomm-messaging/ns/v2"
  ]
```

We will need to add to the W3C DID Registry the following entry for type DIDCommMessaging:
- Normative Reference: https://identity.foundation/didcomm-messaging/spec
- JSON-LD: https://identity.foundation/didcomm-messaging/ns/v2

@tplooker @OR13 any JSON-LD recommendations? What else would be needed?


OR13:
hosting: 

https://github.com/w3c-ccg/lds-jws2020/blob/master/docs/contexts/v1/index.json
https://w3c-ccg.github.io/lds-jws2020/contexts/v1

redirects:

https://w3id.org/jws/v1
https://w3id.org/security/jws/v1

OR13:
You might consider separating your vocab from your spec, so you will have a nice list of terms:

https://identity.foundation/didcomm-messaging/ns

and then their context:

https://identity.foundation/didcomm-messaging/ns/v2

whatever value you choose for your vocab, may get singed over when you canonicalize in JSON-LD.... so you should make that IRI as permanent and nice as you can.

awoie:
Current proposal is as follows ...

Example in DID Docs:
```
"@context":[
    "https://www.w3.org/ns/did/v1",
    "https://identity.foundation/didcomm-messaging/v2"
  ]
```

or

```
"@context":[
    "https://www.w3.org/ns/did/v1",
    "https://w3id.org/didcomm/v2"
  ]
```

The following redirects will be created for the vocabulary for the following URLs:
- https://identity.foundation/didcomm-messaging
- https://w3id.org/didcomm

... they will be redirected to https://identity.foundation/didcomm-messaging/spec

The following redirects will be created for the contexts for the following URLs:
- https://identity.foundation/didcomm-messaging/v2
- https://w3id.org/didcomm/v2

... they will be redirect to https://identity.foundation/didcomm-messaging/contexts/v2/index.json

@OR13 @tplooker can someone pls review? thanks.

OR13:
I would recommend using `w3id.org` for all long lived documentation... since it can be updated if the spec moves, etc...  we tend to handle this issue like this:

1. register a w3id.org namespace for the work item asap.
2. redirect from the registered namespace to a github pages hosted vocabulary document and JSON-LD context.
3. once the spec / vocab is realtively stable, split the spec and vocab apart... like this: 

- https://www.w3.org/ns/activitystreams
- https://www.w3.org/TR/activitypub/

Notice the namespace is not the same as the spec.

awoie:
I forked the w3id repo into DIF and created a PR here: https://github.com/perma-id/w3id.org/pull/2015

awoie:
@dhh1128 @swcurran @TelegramSam @tplooker I guess we can merge this PR if nobody objects to `type`: `DIDCommMessaging` ?

docs/contexts/v2/index.json
```diff
@@ -0,0 +1,9 @@
+{
+    "@context": [
```
tplooker, line 2: I dont think we need an array here?

awoie: ok, removed the array

tplooker (approved):

awoie:
@TelegramSam @tplooker merge conflicts solved, please merge.

PR #161, opened 2021-02-08
Attachments

TelegramSam:
First attempt at attachments. comments welcome.


general_messaging_constructs.md
```diff
@@ -4,9 +4,75 @@
...
+            "encrypted_to": "",
+            "other_details": "about attachment"
+        }
+    },
+    "attachments": [
+        {
+			"id": "1",
+            "description": "example b64 encoded attachment",
+            "data": {
+            	"base64": "WW91ciBob3ZlcmNyYWZ0IGlzIGZ1bGwgb2YgZWVscw=="
```
OR13, line 60: perhaps we can require this data to be compressed with a registered compression alg, like gzip.... every time i see base64 (url?) I think about inflation...

TelegramSam: Requiring it is further than I want to go, but we should definitely support it.

general_messaging_constructs.md
```diff
@@ -4,9 +4,75 @@
...
+    "to": ["did:example:mediator"],
+    "body":{
+        "attachment_id": "1",
+        "encrypted_details": {
+            "id": "x",
+            "encrypted_to": "",
+            "other_details": "about attachment"
+        }
+    },
+    "attachments": [
```
kdenhartog, line 55: What's the main advantage of keeping the attachment outside the body of the payload? @dhh1128 might be able to point out a few, but I'm not remembering from when this was originally designed.

general_messaging_constructs.md
kdenhartog, line 60: Any reason that we couldn't directly make the value of the varying properties the value of data rather than nesting it inside an object with a property? Then we could specify the mime-type of the data using a type property at the same level of data.

E.g.

```
{
	    "id": "2",
            "description": "example linked attachment",
            "type": "application/didcomm-hash-attachment+json"
            "data": {
            	"hash": "<multi-hash>",
                "links": ["http://path/to/resource"]
        	}
        }
```

Main thing I'm thinking about here is how we generically structure this to make this support any type of data in here.

general_messaging_constructs.md
dhh1128: I think this facilitates a solution to avoid re-encrypting data that's already been encrypted once. @TelegramSam ?

general_messaging_constructs.md
dhh1128: Isn't that what we're doing? I don't see how your (Kyle's) proposal differs from what's in the PR -- or how it differs from the way attachments handled these auxiliary properties in DIDComm v1.

kdenhartog: The type property ends up more explicit which makes signing/hashing of attachment data easier to process. For example, the way in which I produce a hash of a link which is a PDF would be different than the way I would produce a hash of a JSON object available at the link.

Additionally, when it comes to links I think only signing a hash would make things simpler than signing the content at the link directly.

general_messaging_constructs.md
```diff
@@ -4,9 +4,75 @@
...
+### Reference
+
+Attachments are contained within a list in the `attachments` header.
+
+Each attachment is contained within the following structure.
+
+- `id`: Identifies attached content within the scope of a given message. Recommended on appended attachment descriptors. Possible but generally unused on embedded attachment descriptors. Never required if no references to the attachment exist; if omitted, then there is no way to refer to the attachment later in the thread, in error messages, and so forth. Because `id` is used to compose URIs, it is recommended that this name be brief and avoid spaces and other characters that require URI escaping.
+- `description`: An optional human-readable description of the content.
+- `filename`: A hint about the name that might be used if this attachment is persisted as a file. It is not required, and need not be unique. If this field is present and `mime-type` is not, the extension on the filename may be used to infer a MIME type.
+- `mime_type`: Describes the MIME type of the attached content. Optional but recommended.
```
dhh1128, line 30: I suggest we go with "media_type" instead of "mime_type", since this is now the preferred verbiage in IETF RFC-land (see https://tools.ietf.org/html/rfc2046).

general_messaging_constructs.md
kdenhartog: As an example how would I support providing CBOR data using the current design?

At first glance, the easiest way is base64 encode the CBOR data, but after I finish decoding that data how do I know if the decoded data is CBOR/ASCII/etc.

Essentially my intent with using mime-types is to more generically support sending any resource format available on the web today and setup an intuitive path for hashing/signing that data when necessary.

general_messaging_constructs.md
kdenhartog: Wouldn't the attachment data still be placed inside the encryption envelop though since it's a part of the JWM?

general_messaging_constructs.md
dhh1128: @kdenhartog : I have studied what you are proposing, and what Sam proposed, and I am not confident that I understand the difference. Is it just that you're proposing a `type` field that's a sibling of `data`, instead of the `mime_type` field that Sam proposed (that's also a sibling of `data`)?

I don't understand what you mean by "we add a new property each time we want to support a different content type." There's only one core property for all content types -- `base64`. The one exception to this is JSON. Because JSON can be embedded in the message without transformation, I guess we introduced `data.json` -- but I don't think we'd introduce `data.anythingelse`?

>after I finish decoding that data how do I know if the decoded data is CBOR/ASCII/etc.

By looking at the `mime_type` field. Am I missing something?

dhh1128: >The type property ends up more explicit which makes signing/hashing of attachment data easier to process. For example, the way in which I produce a hash of a link which is a PDF would be different than the way I would produce a hash of a JSON object available at the link.

This puzzles me. Why would we want to hash PDFs different from JSON? For any content type, we take the raw content and hash it with a generic algorithm (e.g., SHA3), and we're done. Right?

kdenhartog: TL;DR I think mime-types need to be required and we should be using them to assist with determining the how to canonicalize data in any arbitrary format.

>Is it just that you're proposing a type field that's a sibling of data, instead of the mime_type field that Sam proposed (that's also a sibling of data)?

Sorry, I didn't clearly communicate my misalignment. For one I didn't name it properly. Secondly, the problem I see this as a required property, not an optional one.

> There's only one core property for all content types -- base64. The one exception to this is JSON. Because JSON can be embedded in the message without transformation, I guess we introduced data.json -- but I don't think we'd introduce data.anythingelse? I'd be cheerful about removing data.json so we have no exceptions...

I don't quite see it the same, but that's because I'm interpreting it based on what I'm reading in the document more so at this point. Based on what I'm reading and seeing in the examples, `encrypted`, `hash` + `links` (+ `jws), `base64` (+ `jws`) are all core formats. `json` is also a core format as well at the moment, but I'd agree with dropping it. In any case, the reason we've ended up with multiple formats is that we started to use attachments to pass a variety of different data types inside the messages. So my hypothesis (and the reason for me talking about CBOR - It's an edge case that seemed plausible to introduce a new property into data) is that as we encounter more data formats that don't fit within our core types we're likely to introduce more properties. Instead, what I'm hoping to do with mime-type is leverage it to deterministically identify the content which is passed via the core properties we're setting out to define. However, when mime-type is optional it makes it hard to know if we'll extend for new data formats by adding new core data properties or by reusing the existing ones and distinguishing between the data passed with a mime-type. Furthermore, if we could get down to using only a single core property (e.g. only base64) then I'd like to eliminate the object structure of the `data` value all together and just set the value directly. I'm not sure if this is possible yet.


> This puzzles me. Why would we want to hash PDFs different from JSON? For any content type, we take the raw content and hash it with a generic algorithm (e.g., SHA3), and we're done. Right?

PDFs will canonicalize to the input bytes differently than JSON which will be different than base64. This is especially gnarly when we start dealing with arbitrary data formats available with links. My thinking is that we can use the mime-type to know how to canonicalize the data into bytes in a deterministic way in order to handle this. For example, "application/json" would use JCS, "application/ld+json" might use the RDF canonicalization, "audio/mpeg" available at a link wouldn't need it because it is a byte stream already, and "image/jpeg" at a link might end up canonicalizing to bytes in a completely different way too. In any case, we need some way to know how to produce the hashes of these different types of resources and my proposal is to rely on the mime-type by coupling a method of canonicalization to that mime-type. Where possible we should reuse as much as possible too. For example if you're dealing with base64 data just ascii decode it rather than base64 decode and then canonicalized the decoded resulting data.

general_messaging_constructs.md
kdenhartog: +1 Additionally, let's specify it in a way that references RFC2046

general_messaging_constructs.md
kdenhartog: For reference, my thinking here is influenced by this discussion as well https://github.com/hyperledger/aries-rfcs/issues/556

dhh1128: Thanks, @kdenhartog . I feel much more confident that I understand your perspective now. The extra commentary really helped.

I am happy with the idea that we require a mime type. All types would be handled the same way. (It's not an accident that the very mechanism we're considering here -- MIME -- is the mechanism for attachments in email. So making it a requirement here makes total sense to me.)

I'm also happy to dispense with JSON's special handling. If we're attaching it, it's just another stream of bytes, and it shows up base64 encoded like any other mime type. (The argument to special-case JSON was stronger in DIDComm v1, where an attachment could appear in the body of a message. There, having the ability to read it directly was a somewhat nice convenience. Now that we're talking about moving it out of the body, I don't see any reason why JSON needs to be special.)

As far as flattening the structure (moving fields out from under `data`): I have no strong opinion. I'll be happy with whatever you and others negotiate.

Re. canonicalization: I want to say that canonicalization is out of scope. But I agree after re-reading https://github.com/hyperledger/aries-rfcs/issues/556 that there's a funky interaction with signing algorithms. I like the solution you proposed in your final comment: https://github.com/hyperledger/aries-rfcs/issues/556#issuecomment-712589366. This would allow dumb JWS-style signing of an uncanonicalized stream of bytes (sufficient for most use cases) -- and in the cases where the attachment needs more specialized signing, the byte stream would include the signature and the mime type would tell us how to interpret it.

What do you think of that compromise?

general_messaging_constructs.md
TelegramSam: More on this in today's meeting.

general_messaging_constructs.md
kdenhartog: Just want to state on this thread (said it on the 3/9 call) that I agree with that compromise.

tplooker:
Discussed on the 8/6/21 call, resolved to merge

PR #163, opened 2021-02-15
Update jwm.md

### LINKED ISSUE #132: thread_id implicit relationship to message id semantics
The language for thread_id relationship is stronger than the same Aries RFC's language. This language states that the thread_id must always start out as message id. Instead of the implicit relationship in the Aries RFC where if not specified it is implicitly defined as the message id.

```thread_id - OPTIONAL. Identifier for the current protocol's thread. The thread_id attribute value MUST be equal to the id of the message that initiated the protocol. All messages are part of a thread. If thread_id is not present, the thread for the message is identified by id.```

Question: is this change intentional? 

Suggested change to the language: 
```If the thread_id is defined, the thread_id is the value given. But if the thread_id is not defined, the thread_id is implicitly defined as the id of the given message and that message is the first message of a new thread.```

devin-fisher:
thread_id implicit relationship to message id semantics

fixes #132


devin-fisher:
@TelegramSam Here is the PR we discussed in Monday's meeting.

kdenhartog:
Noting that we're dropping normative statements in this. Could we keep those in place while updating this language?

devin-fisher:
@kdenhartog Does the latest version hit the mark better?

kdenhartog (approved):
Looks good to me. Thanks for the changes.

PR #166, opened 2021-03-01
Add P-384 Curve to must be supported

troyronda:
- Adds `P-384` to the curves that must be supported.
- Adds note to `P-256` that is deprecated in favor of `P-384`.

### DIFF
--- a/encryption.md
+++ b/encryption.md
@@ -9,7 +9,8 @@ For the keys involved in key agreement, the following elliptic curves MUST be su
 | Curve  | Description                                                  |
 | ------ | ------------------------------------------------------------ |
 | X25519 | The underlying curve is actually `Curve25519`, however when used in the context of Diffie-Hellman the identifier of `X25519` is used |
-| P-256  | NIST defined P-256 elliptic curve                            |
+| P-384  | NIST defined P-384 elliptic curve                            |
+| P-256  | NIST defined P-256 elliptic curve - deprecated in favor of P-384 |
 
 For content encryption of the message, the following algorithms MUST be supported.
 


TelegramSam:
Discussed in WG 20210301. Open for feedback for a week.

TelegramSam:
@awoie @tplooker please comment if this causes issues.

bumblefudge:
I, for one, prefer Solange but hey, to each their own
https://malware.news/t/everyone-loves-curves-but-which-elliptic-curve-is-the-most-popular/17657

bumblefudge:
All joking aside I think the eIDAS world and the HSM world would lobby for not dropping 256 altogether or demoting it to a "SHOULD" but can deal with being the last one picked on the schoolyard.

kdenhartog:
This is the biggest problems with default crypto selections. Every government picks something different that MUST be supported and then we're left with a smorgasbord of required options. While I'm generally in favor of this because it makes my life easier (NZ requires P-384 usage rather than P-256), I'm not certain of the impact on others and what it means in terms of them having to get governments to make modifications to legislation.

bumblefudge:
Checked with oliver OOB and he pointed me to [this 5-year old document](https://ec.europa.eu/cefdigital/wiki/download/attachments/82773108/eidas_-_crypto_requirements_for_the_eidas_interoperability_framework_v1.0.pdf) that says P-256 and P-384 are both equally valid for securing TLS to convey eIDAS material.  Insofar as TLS is analogous to a DIDComm channel, and Brussels doesn't make a more explicitly preference or deprecation tomorrow, either should be fine? So false alarm on the eIDAS front, sorry to be a font of misinformation.

awoie:
Yep, LGTM.

kdenhartog:
Sweet thanks for double checking on that front!

kdenhartog (approved):

kdenhartog:
On a side note, has anyone checked support for P-384 in HSMs? My assumption was it's fairly well supported by FIPS compliant HSMs, but that's only an assumption.

PR #167, opened 2021-03-01
Add accept property to indicate media type

troyronda:


TelegramSam:
Discussed WG 20210301

kdenhartog:
Is a JWM style message meant to be used with older Aries RFCs or Will JWM only be used with the new encryption envelope? Also have we thought through what envelopes this would be included in? For example when bootstrapping this is only useful if it's in a signed, but not encrypted message or an OOB message.

I assume these considerations were discussed in the WG call this week, but wanted to check since I wasn't on this week.

tplooker:
Discussed on WG call 23/04/21 resolved to merge

PR #168, opened 2021-03-01
Fix mime type in transports

troyronda:


TelegramSam:
Discussed WG 20210301.

PR #172, opened 2021-03-22
Fixed to/next inconsistencies in the routing section

TelegramSam:
Issue surfaced by Ivan on the Slack channel.


kdenhartog (approved):
Thanks for writing this up.

tplooker:
DIscussed on WG call 23/04/21 resolved to merge

tplooker:
@TelegramSam can you resolve conflicts?

35359595:
Can we keep this one open until next WG meeting? Would like to provide some more clarifications of what's bothering me (findings during implementation).

Here's the  original post from slack:

> Mediated message confusion: Section 9.3 says that field to is used for mediated messages, which is logical. but new addition in section 9,1,4 introduces separate body structure for mediated message, which brings some inconveniences for the implementor and limits mediated message to only one forward recipient. I’ve implemented this re-wrap on Friday and it feels really weird. Why not just use to field in mediated inner (post decryption/verification) message as REQUIRED and body to be the payload to forward to next. This will:
> Allow to have multiple forwards from single mediated envelope (reduce transport);
> Ease parse of mediated envelope as no serialization for body is needed;
> Keep those addresses in the same place for any types of messages instead splitting them into two different locations;
> Preserve payload abstracted from DIDComm and keep all “technical” data in header;
> P.S. there is even a hint on multiple payloads for forward? Why to do it if we have this?

TelegramSam:
> Can we keep this one open until next WG meeting? Would like to provide some more clarifications of what's bothering me (findings during implementation).

We have held this PR open for discussion. If you cannot attend or otherwise further this discussion in today's WG meeting, we will merge it and can resolve any remaining discussion in another PR or in issues.

PR #174, opened 2021-03-29
Encrypted Attachments

TelegramSam:


TelegramSam:
As Discussed in WG call 20210419, we will not be solving the 33% bloat problem at this time.

PR #177, opened 2021-04-05
Explain profiles

dhh1128:

### DIFF
--- a/docs/spec-files/message_structure.md
+++ b/docs/spec-files/message_structure.md
@@ -52,6 +52,26 @@ When persisted as a file or attached as a payload in other contexts, the file ex
 
 ![DIDComm Encrypted Message Icon](../collateral/dcem-128.png)
 
+## Negotiating Compatibility
+
+When parties want to communicate via DIDComm, a number of mechanisms must align. These include:
+
+1. The type of service endpoint used by each party
+2. The key types used for encryption and/or signing
+3. The format of the encryption and/or signing envelopes
+4. The encoding of plaintext messages
+5. The protocol used to forward forward and route
+6. The protocol embodied in the plaintext messages
+
+Although DIDComm allows flexibility in each of these choices, it is not expected that a given DIDComm implementation will support many permutations. Rather, we expect a few sets of choices that commonly go together. We call a set of choices that work well together a __profile__. Profiles are identified by a string that matches the conventions of IANA media types, but they express choices about plaintext, encryption, signing, and routing in a single value. The following profile identifiers are defined in this version of the spec:
+
+### Defined Profiles
+
+* `didcomm/aip1`: The encryption envelope, signing mechanism, plaintext conventions, and routing algorithms embodied in Aries AIP 1.0, circa 2020.
+* `didcomm/aip2`: The encryption envelope, signing mechanism, plaintext conventions, and routing algorithms embodied in Aries AIP 2.0, circa 2020.
+* `didcomm/v2`: The encryption envelope, signing mechanism, plaintext conventions, and routing algorithms embodied in this spec.
+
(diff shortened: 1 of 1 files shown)


dhh1128:
@TelegramSam and @troyronda: I don't know if this is the right place to put this content, and I also don't know how much what I added here overlaps with what Troy will write. But here's a quick attempt to capture what we discussed on the call today. To be refined and/or replaced with something better...

docs/spec-files/message_structure.md
```diff
@@ -52,6 +52,26 @@ When persisted as a file or attached as a payload in other contexts, the file ex
...
+4. The encoding of plaintext messages
+5. The protocol used to forward forward and route
+6. The protocol embodied in the plaintext messages
+
+Although DIDComm allows flexibility in each of these choices, it is not expected that a given DIDComm implementation will support many permutations. Rather, we expect a few sets of choices that commonly go together. We call a set of choices that work well together a __profile__. Profiles are identified by a string that matches the conventions of IANA media types, but they express choices about plaintext, encryption, signing, and routing in a single value. The following profile identifiers are defined in this version of the spec:
+
+### Defined Profiles
+
+* `didcomm/aip1`: The encryption envelope, signing mechanism, plaintext conventions, and routing algorithms embodied in Aries AIP 1.0, circa 2020.
+* `didcomm/aip2`: The encryption envelope, signing mechanism, plaintext conventions, and routing algorithms embodied in Aries AIP 2.0, circa 2020.
```
troyronda, line 71: There are actually two flavors of AIP2 - one with Envelope v1 (RFC 19) and one with Envelope v2 (RFC 587).

Currently the media types in the Aries accept properties are:
- application/json;flavor=didcomm-msg (roughly the equivalent of AIP1).
- application/didcomm-encrypted+json;cty=application/json (roughly the equivalent of AIP2 with Envelope v2).

With AIP2 + Envelope v2, we are allowing an AIP to enable community transition with an AIP2 target. If we could include that target concept here, then we could propose the same profile media type concept in Aries RFCs.

dhh1128: Agreed. Good point.

What if we did the following:

* `didcomm/aip1`: (no change in descrip)
* `didcomm/aip2retro`: The signing mechanism, plaintext conventions, and routing algorithms embodied in Aries AIP 2.0, circa 2020 -- with the old-style encryption envelope from Aries RFC 0019. This legal variant of AIP 2.0 minimizes differences with codebases that shipped AIP 1.0 support.
* `didcomm/aip2pro`: The signing mechanism, plaintext conventions, and routing algorithms embodied in Aries AIP 2.0, circa 2020 -- with the new-style encryption envelope from Aries RFC 0587. This legal variant of AIP 2.0 lays the foundation for DIDComm v2 support by anticipating the eventual envelope change.
* `didcomm/v2` (keep current description)

troyronda: Thanks @dhh1128 - nice writeup.

dhh1128: I've updated the text to address this point. What do you think, @troyronda ?

kdenhartog (approved):

docs/spec-files/message_structure.md
```diff
@@ -52,6 +52,27 @@ When persisted as a file or attached as a payload in other contexts, the file ex
...
 
+## Negotiating Compatibility
+
+When parties want to communicate via DIDComm, a number of mechanisms must align. These include:
+
+1. The type of service endpoint used by each party
+2. The key types used for encryption and/or signing
+3. The format of the encryption and/or signing envelopes
+4. The encoding of plaintext messages
+5. The protocol used to forward forward and route
```
troyronda, line 63: Typo - two forwards

troyronda (approved):

docs/spec-files/message_structure.md
dhh1128: Good catch. Updated.

PR #195, opened 2021-05-17
Anoncrypt warning

TelegramSam:


docs/spec-files/encryption.md
```diff
@@ -10,7 +10,9 @@ For an encrypted DIDComm Message, the JWA of `ECDH-1PU` defined by [draft](https
 
 ## Anonymous Encryption
 
- When a sender would like to encrypt a message for a recipient or multiple recipients but not be authenticated by the recipients as the party who encrypted the message, the JWA of `ECDH-ES` defined by [RFC 7518](https://tools.ietf.org/html/rfc7518#section-4.6) SHOULD be used within the structure of a JWE.
+When a sender would like to encrypt a message for a recipient or multiple recipients but not be authenticated by the recipients as the party who encrypted the message, the JWA of `ECDH-ES` defined by [RFC 7518](https://tools.ietf.org/html/rfc7518#section-4.6) SHOULD be used within the structure of a JWE.
+
+Anonymous Encryption removes authenticated encryption, a significant benefit of the DIDComm specification. Use of Anonymous Encryption MUST NOT be paired with a method of message authentication other than Authenticated Encryption as defined in this specification.
```
troyronda, line 15: It would be nice to tighten this statement to only refer to the overall message authentication as it relates to DIDComm itself. Although it seems a bit unnecessary as DIDComm doesn't define a different message authentication?

Outside of the DIDComm spec aspects, I assume there is some implementation guidance but that wouldn't be testable / MUSTable.

OR13, line 15: I am not sure what this means... can we try for simpler language here?

OR13: also, this statement would appear to be untestable, which is sometimes a sign that it might be better to reward it.

OR13: maybe better to say: "be careful trusting content submitted anonymously without signatures from parties you trust?"

PR #198, opened 2021-06-15
fixes #165 - updated typ/cty language for JWS/JWE/Plain

### LINKED ISSUE #165: Require always `cty: "JWM"`
The JWM draft's use of [`cty`](https://tools.ietf.org/html/draft-looker-jwm-01#section-4.2) follows that of [RFC7519](https://tools.ietf.org/html/rfc7519#section-5.2), [RFC7516](https://tools.ietf.org/html/rfc7516#section-4.1.12), and [RFC7515](https://tools.ietf.org/html/rfc7515#section-4.1.10), which is to make this parameter optional/NOT RECOMMENDED.

I propose that for DIDComm v2 this header MUST be present (and its value MUST be `JWM` just like the spec says right now).

Doing this allows a cleaner transition V1 -> V2.

Edit: we could specify a separate `cty` for "regular" (ie non-nested) JWMs so as not to clash with the nested ones, which require `JWM`.

awoie:
Potentially fixes #165 

* removed `typ` from encryption.md (seemed redundant).
* added explicit language that JWE/JWS are JWM.
* added `typ` as mandatory header for didcomm plaintext messages. 
* made `didcomm-signed+json`/`didcomm-encrypted+json` mandatory for JWE/JWS for `typ` property.
* updated and fixed OOB examples.

Overview (updated - including resolution from DIDComm WG meeting June 21st; removed language on `cty` which might need a separate issue if still required):

| DIDComm message | `typ` |
|-----------------|-------|
| Encrypted | `application/didcomm-encrypted+json`|
| Signed, then encrypted | `application/didcomm-encrypted+json`|
| Encrypted, then encrypted | `application/didcomm-encrypted+json` |
| Signed | `application/didcomm-signed+json` |
| Plaintext | `application/didcomm-plain+json`|

### DIFF
--- a/docs/spec-files/encryption.md
+++ b/docs/spec-files/encryption.md
@@ -46,12 +46,6 @@ TODO: Include language about safe nonce considerations.
 | ECDH-1PU+A256KW | P-521           | EC                 | ECDH-1PU key wrapping using key with NIST defined P-521 elliptic curve to create a 256 bits key as defined in [ecdh-1pu](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-04#section-2) |
 | ECDH-1PU+A256KW | X25519          | OKP                | ECDH-1PU X25519 ([RFC7748 section 5](https://tools.ietf.org/html/rfc7748#section-5)) to create a 256 bits key as defined in [ecdh-1pu](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-04#section-2) |
 
-## Media Types
-
-The media type of the envelope MUST be set in the `typ` [property](https://tools.ietf.org/html/rfc7516#section-4.1.11) of the JWE and the media type of the payload MUST be set in the `cty` [property](https://tools.ietf.org/html/rfc7516#section-4.1.12) of the JWE. The [message types](#message-types) section provides general discussion of the media types.
-
-For example, following the guidelines of [message types](#message-types), an encrypted envelope with a plaintext DIDComm payload contains the `typ` property with the value `application/didcomm-encrypted+json` and `cty` property with the value `application/didcomm-plain+json`.
-
 ## Perfect Forward Secrecy
 
 Due to the triple Key Derivation algorithm used in ECDH-1PU, all messages sent via DIDComm have weak perfect forward secrecy without any additional security added by the transport layer. In ECDH-1PU this is achieved by encrypting the content encryption key with the output of the hash of the Ze (ECDH of ephemeral key and recipient static key) and Zs (ECDH of static sender key and recipient static key). With Ze bringing the changed derived secret in each message and Zs bringing the repudiable authenticity of each message, the resulting Z (hash of Ze and Zs) carries the properties of weak perfect forward secrecy and repudiable authenticity for each message as well.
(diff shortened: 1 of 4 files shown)


awoie:
We might get rid of `typ` in the DIDComm plaintext message but then we would need to do content sniffing or assume that every DIDComm plaintext message is registered in didcomm.org if we want to use the `type` header.

awoie:
@Baha-sk also pinging Baha since he created #165

Baha-sk:
This needs to be aligned with Aries-RFC 587 [Media type](https://github.com/hyperledger/aries-rfcs/tree/master/features/0587-encryption-envelope-v2#media-type) section in order to align Interop between AIP2.0 and DIDComm V2.

Also, I'm in favour with @dhh1128's [comment](https://github.com/decentralized-identity/didcomm-messaging/issues/165#issuecomment-796152914) about not using JWM as content type as it doesn't convey the true content type of the payload.

awoie:
> This needs to be aligned with Aries-RFC 587 [Media type](https://github.com/hyperledger/aries-rfcs/tree/master/features/0587-encryption-envelope-v2#media-type) section in order to align Interop between AIP2.0 and DIDComm V2.
> 
> Also, I'm in favour with @dhh1128's [comment](https://github.com/decentralized-identity/didcomm-messaging/issues/165#issuecomment-796152914) about not using JWM as content type as it doesn't convey the true content type of the payload.

Yep, happy to drop JWM but then the JWM spec needs to change since it is required for nested JWMs.

kdenhartog:
> > This needs to be aligned with Aries-RFC 587 [Media type](https://github.com/hyperledger/aries-rfcs/tree/master/features/0587-encryption-envelope-v2#media-type) section in order to align Interop between AIP2.0 and DIDComm V2.
> > Also, I'm in favour with @dhh1128's [comment](https://github.com/decentralized-identity/didcomm-messaging/issues/165#issuecomment-796152914) about not using JWM as content type as it doesn't convey the true content type of the payload.
> 
> Yep, happy to drop JWM but then the JWM spec needs to change since it is required for nested JWMs.
> 
> **Updated:**
> 
> The PR uses `application/JWM` only for nested JWMs. I'm supportive of having always (also for nested structures) the actual media type in `cty` but since the outer envelope always has `application/didcomm-...` as `typ`, you would at least be able to start DIDComm processing.
> 
> Meaning...
> 
>     1. check if `typ` equals `application/didcomm-encrypted+json`
> 
>     2. decrypt encrypted payload
> 
>     3. check if decrypted message is a potential DIDComm message (e.g., `cty` equals `application/JWM`, `application/didcomm-...`, ...)
> 
>     4. check if `typ` of decrypted message is either `didcomm-encrypted+json`, `didcomm-signed+json` or `didcomm-plain+json`.
> 
>     5. and so on ...

I think this is the right way to go. When referring to a didcomm based JWM I think the more specific `application/didcomm-plain+json` makes heaps more sense then the more generic `application/JWM` which I believe is what @dhh1128 had originally pointed out.

awoie:
Update from DIDComm meeting (Jan 21): 
- keep typ property language
- defer cty language until @kdenhartog is available to discuss

awoie:
> Update from DIDComm meeting (Jan 21):
> 
> * keep typ property language
> * defer cty language until @kdenhartog is available to discuss

@kdenhartog @TelegramSam @dhh1128 I updated the PR to remove language on `cty` and only kept language on usage `typ`. We can create a separate ticket for `cty` if needed.

kdenhartog (approved):
In general I think what's described within this PR looks good. We've come across a similar problem that relies upon what we do here, but rather than holding it up to discuss a largely orthogonal problem I'll move it to a separate issue #220 for us to discuss how we'd use this to help solve it.

PR #200, opened 2021-06-16
Propose threading section

dhh1128:


TimoGlastra (approved):
Looks great!

If I may make a suggestion, I think it will be nice to add a code/json example of the `id`, `thid`, `pthid` properties being used.

OR13 (approved):
this is a lot :)

docs/guide-files/threading.md
```diff
@@ -0,0 +1,33 @@
+## Threading
+
+DIDComm's threading mechanism enables referencing individual messages, the sets of messages that constitute a logically independent interaction (a "run" or "instance" of a protocol), or even the hierarchies of messages that model nested and recursive workflows. This helps with troubleshooting, debugging, error reporting, resend logic, interactions across more than two parties, and patterns of communication that are more flexible or asynchronous than simple request+response.
+
+Some context about the intent and best practices for these features may be helpful:
+
+* For cybersecurity reasons, the `id` properties of related messages SHOULD NOT bear any resemblance to one another; `id` values SHOULD be opaque and SHOULD NOT be used to "tunnel" other semantics. This lack of a detectable relationship should be characteristic of `id` values for messages that stand in sequence to one another, as well as the `id` properties of `forward` messages seen by mediators vis-a-vis the `id` of the innermost plaintext messages they carry.
```
kdenhartog, line 7:
```suggestion
* For cybersecurity reasons, the `id` properties of related messages are intended not to bear any resemblance to one another; `id` values are expected to be opaque and will likely cause additional issues if they include overloaded semantics that are unique to a particular implementation. This lack of a detectable relationship should be characteristic of `id` values for messages that stand in sequence to one another, as well as the `id` properties of `forward` messages seen by mediators vis-a-vis the `id` of the innermost plaintext messages they carry.
```

I think it makes sense to remove the normative statements here and keep this as implementation guidance. Here's a suggestion on how we can clean it up to make that happen.

docs/guide-files/threading.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Threads
+
+Like threads in email, a **thread** in DIDComm intends to model a discrete interaction. For protocols that have a beginning and end (e.g., exchanging credentials, playing a game of chess, making a payment), a DIDComm thread maps exactly onto the sequence of messages that embody one such interaction; when a new interaction like this starts, the parties MUST impute a new thread to the context.
```
kdenhartog, line 11:
```suggestion
Like threads in email, a **thread** in DIDComm intends to model a discrete interaction. For protocols that have a beginning and end (e.g., exchanging credentials, playing a game of chess, making a payment), a DIDComm thread maps exactly onto the sequence of messages that embody one such interaction; when a new interaction like this starts, the parties are expected to impute a new thread to the context.
```

I was considering suggesting we move this section to the spec files because that normative statement seemed useful. However, I wasn't certain how we'd test that so opted to change this to non normative text. If you know of a way that we could make this testable, I'm open to moving this particular text back to the spec and keeping this normative.

docs/guide-files/threading.md
```diff
@@ -0,0 +1,33 @@
...
+
+However, some DIDComm protocols are not obviously discrete. An example of this might be a human-friendly chat; humans don't necessarily perceive crisp interaction boundaries around any given subset of the messages they trade back and forth. In such cases, a thread might map to a topic, a unit of time, or something else. Applications MAY structure the thread construct in whatever way makes sense to the implementers, but MUST describe the threading assumptions of the protocol in the protocol's documentation.
```
kdenhartog, line 13:
```suggestion
However, some DIDComm protocols are not obviously discrete. An example of this might be a human-friendly chat; humans don't necessarily perceive crisp interaction boundaries around any given subset of the messages they trade back and forth. In such cases, a thread might map to a topic, a unit of time, or something else. Applications can structure the thread construct in whatever way makes sense to the implementers, but MUST describe the threading assumptions of the protocol in the protocol's documentation.
```

This last MUST in here is an interesting one. I think we may actually want to move this back to the spec even though it's not machine testable. This is very similar to what we're doing with defining how to author a DID Method with DID Core.

docs/guide-files/threading.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Parent Threads
+
+Parent threads and child threads do NOT necessarily represent *threads of execution* or concurrency; they only represent logically separable *sequences of messages*. A parent thread MAY trigger multiple child threads, and MAY cascade to multiple levels of nested message threads (for example, issuing a credential triggers payment, which in turn triggers feature discovery). Parent threads and child threads do not have a simple containment or ownership relationship. By definition, a parent is only a trigger &mdash; not necessarily a controller or stakeholder. A parent thread MAY end before or after any of its child threads. Parent and child threads MAY interact or share data, but are not required to do so. (How they do so is out of scope in this spec, but see [Coprotocols](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0478-coprotocols/README.md) for one way to approach this.)
```
kdenhartog, line 17:
```suggestion
Parent threads and child threads do not necessarily represent *threads of execution* or concurrency; they only represent logically separable *sequences of messages*. A parent thread can trigger multiple child threads which could cascade to multiple levels of nested message threads (for example, issuing a credential triggers payment, which in turn triggers feature discovery). Parent threads and child threads do not have a simple containment or ownership relationship. By definition, a parent is only a trigger &mdash; not necessarily a controller or stakeholder. A parent thread has no guarantees about it's lifecycle and could end before or after any of its child threads. Parent and child threads also can interact or share data, but are not required to do so. (How they do so is out of scope in this spec, but see [Coprotocols](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0478-coprotocols/README.md) for one way to approach this.)
```

kdenhartog: This is very beneficial detail that is going to be helpful for implementers. Thanks for making sure this gets written down.

docs/guide-files/threading.md
```diff
@@ -0,0 +1,33 @@
...
+
+#### Parent and Child Examples
+
+Suppose a DIDComm-based protocol (and therefore, a thread of messages) is underway in which an issuer wants to give a credential to a holder. At a particular stage in this interaction, perhaps the issuer asks the prospective holder of the credential to pay for what they're about to receive. For composability, encapsulation, reusability, and versioning reasons, negotiating and consummating payment is best modeled as a separable interaction from credential exchange &mdash; so a new sequence of messages (dedicated to payment) begins. In this example, the credential issuance interaction (message thread 1) is the parent of the payment interaction (message thread 2). The first message in thread 2 MUST contain a `pthid` header that references the `thid` from thread 1.
```
kdenhartog, line 21: This paragraph looks like it fits better back in the spec text. That normative statement seems important and I think it should remain rather than allowing it to stay in the non-normative guide.

docs/guide-files/threading.md
```diff
@@ -0,0 +1,33 @@
...
+
+>NOTE: When a child protocol is a simple two-party interaction, mentioning the `pthid` in the first message of the child interaction is enough to establish context. However, in protocols involving more than two parties, the first message of the child protocol may not be seen by everyone, so simply mentioning `pthid` once may not provide enough context. Therefore, the rule is that each party in a child protocol MUST learn the identity of the parent thread via the first child protocol message they see. The simplest way to ensure this is to mention the `pthid` with every message in the child protocol.
```
kdenhartog, line 23:
```suggestion
>NOTE: When a child protocol is a simple two-party interaction, mentioning the `pthid` in the first message of the child interaction is enough to establish context. However, in protocols involving more than two parties, the first message of the child protocol may not be seen by everyone, so simply mentioning `pthid` once may not provide enough context. Therefore, the rule is that each party in a child protocol are expected to learn the identity of the parent thread via the first child protocol message they see. The simplest way to ensure this is to mention the `pthid` with every message in the child protocol. 
```

docs/guide-files/threading.md
```diff
@@ -0,0 +1,33 @@
...
+
+>NOTE: When a child protocol is a simple two-party interaction, mentioning the `pthid` in the first message of the child interaction is enough to establish context. However, in protocols involving more than two parties, the first message of the child protocol may not be seen by everyone, so simply mentioning `pthid` once may not provide enough context. Therefore, the rule is that each party in a child protocol MUST learn the identity of the parent thread via the first child protocol message they see. The simplest way to ensure this is to mention the `pthid` with every message in the child protocol. 
+
+Creating a connection is another classic interaction that lends itself to a parent-child paradigm. Alice and Bob rarely connect just to connect; usually they connect so they can conduct some type of business. In such situations, the connection protocol is a parent thread, and whatever business they then conduct becomes a child.
```
kdenhartog, line 25: This part of the section and lower seems like it would make sense to leave it in this implementation guide. It's good detail that helps protocol designers/implementers that should be documented, but the lack of a normative statement lends it to suggest it's probably best not to put back in the spec. For this reason I think it should remain here.

docs/guide-files/threading.md
```diff
@@ -0,0 +1,33 @@
...
+
+Sometimes the parent of a thread is not known, or is so independent of its child as to be irrelevant. If Alice and Bob connected three years ago, each new interaction that they begin doesn't need to declare a `pthid` that references the original interaction that connected them. Declare a parent relationship for a thread when it is likely to be useful.
+
+### DIDComm Message URIs
+
+The `id`, `thid`, and `pthid` properties of any DIDComm message may be combined to form a URI that uniquely identifies the message (e.g., in debuggers, in log files, in archives). When this is further combined with [JSPath](https://github.com/dfilatov/jspath#quick-example), individual elements of messages may also be hyperlinked. See [Linkable Message Paths](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0217-linkable-message-paths/README.md). Such a scheme is out of scope for this spec, and support for it is NOT required of implementers. However, this spec reserves the `didcomm://` URI prefix for future work with semantics like these.
```
kdenhartog, line 33: I think most of this paragraph should remain here, but that semantic reservation probably needs to go back into the spec for greater awareness about the topic even if we may leave it as non-normative. Take a look at [this note](https://w3c.github.io/did-core/#did-url-syntax) in did-core about how the `;` charter was "reserved" without being normative.

kdenhartog (changes requested):
Lots of editorial suggestions around where things should go and some quibbles about normative aspects. In general though, this content is incredibly useful and with some quick back and forth on this I think we can get this merged rather quickly. Thanks for writing all of this detail up!

docs/spec-files/threading.md
```diff
@@ -0,0 +1,66 @@
+## Threading
+
+DIDComm provides a mechanism for referencing individual messages, the sets of messages that constitute a logically independent interaction (a "run" or "instance" of a protocol), or even the hierarchies of messages that model nested and recursive workflows. This helps with troubleshooting, debugging, error reporting, resend logic, interactions across more than two parties, and patterns of communication that are more flexible or asynchronous than simple request+response. It is the foundation for extremely powerful protocol features.
+
+### Message IDs
+
+All plaintext DIDComm messages MUST have an `id` property, declared in the JWM header. A message without an `id` property SHOULD be considered invalid and SHOULD be rejected; it MUST NOT be interpreted as part of a multi-message interaction.
```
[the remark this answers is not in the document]
dhh1128: If we have a one-way, one-message protocol (e.g., something that sends notifications only), I don't think that threading is needed. Therefore, I put "SHOULD" in rather than MUST. Changing this to a MUST makes sense for two-way, multi-step interactions, but I don't think it does for one-way.

docs/spec-files/threading.md
```diff
@@ -0,0 +1,29 @@
...
+
+The value of an `id` property SHOULD be globally, universally unique; at the very least, it MUST be unique across all interactions visible to the set of parties that see a given set of interactions.
+
+### Threads
+
+A thread is uniquely identified by a thread ID. The thread ID is communicated by including a `thid` header in the JWM plaintext of the message. The value of `thid` MUST conform to the same constraints as the value of `id`. The DIDComm plaintext message that begins a thread MAY declare this property for the new thread. If no `thid` property is declared in the first message of an interaction, the `id` property of the message is used as `thid` as well; that is, the message is interpreted as if both `id` and `thid` were present, containing the same value.
```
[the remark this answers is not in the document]
dhh1128: I agree with the sentiment, but I believe the wording you've suggested is backward. If no `thid` property is present, it is the `id` property, not the `thid` property, that plugs the gap. So we can either reword it as:

>If no `thid` property is declared in the first message of an interaction, the `thid` property value must be derived from (and equal to) the `id` property, which is guaranteed to be present

OR

>If no `thid` property is declared in the first message of an interaction, the `id` property of the message MUST be treated as the value of the `thid` as well

I have updated the text to use the second form.

If you disagree about this reversal of the verbiage, please reopen the comment.

docs/spec-files/threading.md
```diff
@@ -0,0 +1,29 @@
...
+
+All subsequent messages in a thread MUST include a `thid` header that contains the same value as the `thid` set in the first message of the thread. Messages that do not share the same `thid` MUST NOT be imputed to the same thread.
```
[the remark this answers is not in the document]
dhh1128: Agree that this is better. Updated.

docs/spec-files/threading.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Parent Threads
+
+When one interaction triggers another, the first interaction is called the **parent** of the second. This MAY be modeled by incorporating a `pthid` header in the JWM plaintext of the child. The value of the child's `pthid` header MUST obey the same constraints as `thid` and `id` values.
+
+### DIDComm Message URIs
+
+The `id`, `thid`, and `pthid` properties of any DIDComm message may be combined to form a URI that uniquely identifies the message (e.g., in debuggers, in log files, in archives). When this is further combined with [JSPath](https://github.com/dfilatov/jspath#quick-example), individual elements of messages may also be hyperlinked. See [Linkable Message Paths](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0217-linkable-message-paths/README.md). Such a scheme is out of scope for this spec, and support for it is NOT required of implementers. However, this spec reserves the `didcomm://` URI prefix for future work with semantics like these.
```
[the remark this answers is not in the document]
dhh1128: I did some research back in the day, and concluded that JSON Pointer was flawed in some way (it might be the number and quality of the impls across programming langs) -- but I don't remember the details. So I have a suspicion that we won't end up using JSON Pointer. However, that doesn't matter here; we can reference both ideas without specifying which we will use.

Updating to address the suggestion.

docs/guide-files/threading.md
dhh1128: Moved back to spec, per suggestion.

docs/guide-files/threading.md
dhh1128: Moved back into spec with previous paragraph, and kept the MUST. I don't think the behavior described with "MUST" can be optional. But let's discuss further if needed.

docs/guide-files/threading.md
dhh1128: What would be the "semantic reservation" part that you'd move back?

docs/guide-files/threading.md
dhh1128: Updated per suggestion.

docs/spec-files/threading.md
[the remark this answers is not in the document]
kdenhartog: Agreed, this change doesn't make sense now. Let's go with what you have.

docs/spec-files/threading.md
```diff
@@ -0,0 +1,42 @@
...
+
+### DIDComm Message URIs
+
+The `id`, `thid`, and `pthid` properties of any DIDComm message may be combined to form a URI that uniquely identifies the message (e.g., in debuggers, in log files, in archives). When this is further combined with something like [JSPath](https://github.com/dfilatov/jspath#quick-example) or [JS Pointer](https://datatracker.ietf.org/doc/html/rfc6901), individual elements of messages may also be hyperlinked. See [Linkable Message Paths](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0217-linkable-message-paths/README.md). Such a scheme is out of scope for this spec, and support for it is NOT required of implementers. However, this spec reserves the `didcomm://` URI prefix for future work with semantics like these.
```
kdenhartog, line 38: Looks like we've doubled up this section in both the guide files and the spec-files. To answer your question about reservation here's a block of suggestion text for the spec files and I've also added a suggestion block for the guide files.

```suggestion
The `id`, `thid`, and `pthid` properties of any DIDComm message may be combined to form a URI that uniquely identifies the message (e.g., in debuggers, in log files, in archives). Such a scheme is out of scope for this spec, and support for it is OPTIONAL for implementers. This spec reserves the `didcomm://` URI prefix for future work with semantics like these and SHALL NOT be used for other purposes.
```

docs/guide-files/threading.md
kdenhartog:
```suggestion
It's generally considered this functionality will be combined with [JSPath](https://github.com/dfilatov/jspath#quick-example), [JS Pointer](https://datatracker.ietf.org/doc/html/rfc6901), or individual elements of messages may also be hyperlinked. For a good idea of the concepts that we'd like to enable with this capability, see [Linkable Message Paths](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0217-linkable-message-paths/README.md).
```

docs/spec-files/threading.md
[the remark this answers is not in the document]
dhh1128: Should we add a note explaining why it's only "SHOULD"? Or maybe, put in "MUST" for any multi-step interaction?

docs/guide-files/threading.md
dhh1128: Okay, I took this out of the guide and put it back into the spec, with verbiage that tracks what you suggested here. You are attributed as author of the commit.

docs/spec-files/threading.md
dhh1128: Okay, I think I correctly accepted both your suggestions. See if you are satisfied. (You are attributed as the author.)

docs/spec-files/threading.md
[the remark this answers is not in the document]
kdenhartog: I think it's fine to leave it as just "SHOULD" without explicit mention. Typically when I read documents these days "SHOULD" means to me "It's done this way unless you got a reason to not do it" which relies upon the reader to be able to come up with (and then they know they're making an exception to a soft rule).

docs/spec-files/threading.md
```diff
@@ -0,0 +1,42 @@
...
+```jsonc
+{
+  "id": "new-uuid-for-payment-thread",
+  "pthid": "id-of-old-credential-issuance-thread"
+  // ...more headers and message content...
+}
+
+```
+
+>NOTE: When a child protocol is a simple two-party interaction, mentioning the `pthid` in the first message of the child interaction is enough to establish context. However, in protocols involving more than two parties, the first message of the child protocol may not be seen by everyone, so simply mentioning `pthid` once may not provide enough context. Therefore, the rule is that each party in a child protocol MUST learn the identity of the parent thread via the first child protocol message they see. The simplest way to ensure this is to mention the `pthid` with every message in the child protocol.
```
kdenhartog, line 34:
```suggestion
When a child protocol is a simple two-party interaction, mentioning the `pthid` in the first message of the child interaction is enough to establish context. However, in protocols involving more than two parties, the first message of the child protocol may not be seen by everyone, so simply mentioning `pthid` once may not provide enough context. Therefore, the rule is that each party in a child protocol MUST learn the identity of the parent thread via the first child protocol message they see. The simplest way to ensure this is to mention the `pthid` with every message in the child protocol.
```

Let's pull this out of the note since that normative statement is needed. From my understanding, note's are intended to be non-normative (in W3C) so shouldn't contain normative language.

dhh1128: Agreed. Done.

docs/guide-files/threading.md
kdenhartog: Looks like this and 2 other comments in this file are the only ones left to resolve.

docs/guide-files/threading.md
kdenhartog: Looks like this and 2 other comments in this file are the only ones left to resolve.

docs/guide-files/threading.md
kdenhartog: Looks like this and 2 other comments in this file are the only ones left to resolve.

docs/spec-files/threading.md
```diff
@@ -0,0 +1,29 @@
...
+
+### DIDComm Message URIs
+
+The `id`, `thid`, and `pthid` properties of any DIDComm message may be combined to form a URI that uniquely identifies the message (e.g., in debuggers, in log files, in archives). When this is further combined with [JSPath](https://github.com/dfilatov/jspath#quick-example), individual elements of messages may also be hyperlinked. See [Linkable Message Paths](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0217-linkable-message-paths/README.md). Such a scheme is out of scope for this spec, and support for it is NOT required of implementers. However, this spec reserves the `didcomm://` URI prefix for future work with semantics like these.
+
+### Gaps, Resends, and Sophisticated Ordering
+
+Message IDs and threads can be used to build very powerful features for detecting missing and out-of-order messages &mdash; and to recovery from them. For more information, see the [Implementers Guide](http://example.com/TODO--find-right-way-to-link-to-section-in-guide).
```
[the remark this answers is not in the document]
kdenhartog: This looks good now.

kdenhartog:
@dhh1128 it looks like 3 of the comments got hidden by github in the review process. Once we figure out what we want to do with them I think this is good to go. Here's links to them:

https://github.com/decentralized-identity/didcomm-messaging/pull/200#discussion_r655076518
https://github.com/decentralized-identity/didcomm-messaging/pull/200#discussion_r655077033
https://github.com/decentralized-identity/didcomm-messaging/pull/200#discussion_r655077554

PR #206, opened 2021-06-18
Add info about advanced sequencing (out-of-order and gap detection, resend)

dhh1128:
>NOTE: when evaluating the diff for this PR, keep in mind that the PR comes from a branch off of the one behind PR #200. Thus, most of the stuff in this PR is redundant with 200. The only part of this PR that's truly novel is the "Advanced Sequencing" section at the bottom of `guide-files/threading.md` -- everything else is redundant and will drop out if merge PR 200.

I'm ambivalent about whether this is the right way to handle the information that the PR embodies. We could also move that info into a DIDComm extension.

The basic problem I'm trying to address with this is that without a discussion of this issue and its solutions, DIDComm's story about tremendous flexibility (many parties, many transports, etc) breaks down a bit in practice. If you're not doing request/response, and if you're using transports that are highly asynchronous, the chances that messages could arrive out of order or that some messages could be dropped entirely is fairly high. So I think the problem is important to a mature DIDComm.

However, we don't currently have any protocols that require this level of sophistication, so a case could be made for this to not make the spec in any form at all. I don't want to burden implementers with stuff that will be irrelevant...

The compromise position that I'm tentatively feeling okay about is to put this info in the Implementers Guide. But making it an extension might be better?

Also, I'm not sure this is the "right" way to support these features. Some alternatives we might want to discuss include:

1. Make the value of `id` be a monotonically increasing integer. This would mean some gaps (the ones where a later message arrives without its predecessors) could be detected without extra headers. However, it would not allow detection of "trailing gaps" (where Alice has sent 7 messages, but Bob only knows about 4 of them).
2. Combine #1 with ACKs or with a header that tells other parties what the last seen `id` was.

Let's discuss on a community call.


dhh1128:
Superseded by #209

PR #209, opened 2021-06-21
Reframe advanced sequencing as an extension

dhh1128:

This supersedes PR #206; it is essentially the same content except that it reframes the mechanism as an extension rather than something described only in the implementer's guide.

### DIFF
--- a/extensions/advanced_sequencing/main.md
+++ b/extensions/advanced_sequencing/main.md
@@ -1,0 +1,34 @@
+# DIDComm Advanced Sequencing Extension
+
+## Summary 
+
+In a simple interaction between two parties that consists of a request and a response, the message types usually imply the order without ambiguity. However, some protocols have more complex requirements. For example, an auction protocol that consists of `bid` messages might produce different outcomes depending on whether a lower bid arrives before or after a higher bid. When such protocols unfold over asynchronous transports that don't guarantee a delivery order, and when more than two participants are involved, understanding the intended sequence requires extra data.
+
+## Scope
+
+This extension does not specify how gaps or out-of-order messages should be *handled* -- only how they should be *detected*. That is because some protocols may consider out-of-order messages harmless, whereas other protocols may be highly sensitive to problems triggered by such conditions. Thus, the scope of this extension is *detecting sequencing issues*. Any protocol that cares about sequencing issues SHOULD use this mechanism as its foundational semantics for understanding the problem, and MAY choose to react to the issues in whatever way the protocol designers deem appropriate. 
+
+## Reference
+
+DIDComm messages that support this extension offer three headers related to sequencing: `sender_order`, `sent_count`, and `received_orders`. These headers are only relevant to certain protocols. Protocol definitions MUST declare whether they use any of these headers. This is because general implementations of DIDComm are not required to implement support for the headers unless/until they intend to support protocols that use them.
+
+### `sender_order`
+
+The `sender_order` header tells other parties how the sender of a message perceives their own ordering. Its value is a monotonically increasing natural/ordinal number (1..<var>N</var>) that tells how many different messages the sender has sent in the current thread. When Alice and Bob are both bidding in an auction protocol, each of them marks their first bid with `sender_order: 1`, their second bid with `sender_order: 2`, and so forth. This allows the auctioneer to detect if Alice's bids arrive in a different order than she intended. It also means that any message can be uniquely identified by its `thid`, the sender, and the value of that sender's `sender_order`. Note how this does NOT clarify the sequence of Alice's messages relative to Bob's.
+
+### `sent_count`
+
+The `sent_count` header enables resend behavior. A resent message contains the same headers (including `id` and `sender_order`) and body as a previous transmission. The first time a message is sent, the `sent_count` for that message is 1, and the header is normally omitted (the value of the header is implicitly 1). The second time the message is sent, the `sent_count` is 2, and the header is added to indicate that the message might be redundant ("this is the second time I've sent this to you"). The header continues to be incremented with each subsequent resend. Once a recipient has received one copy of a message, they MUST ignore subsequent copies that arrive, such that resends are idempotent.
+
+### `received_orders`
+
+The `received_orders` header tells other parties how the sender has experienced the unfolding interaction so far. This allows gaps to be detected, possibly triggering a resend. The value of this header is an array of *gap detector* objects in the format `{"id":<did of party>, "last":<value of biggest sender_order seen from this party>, "gaps": []}`. In our running auction example, if the auctioneer sees bids 1 and 2 from Bob, and bid 1 and 4 from Alice, the auctioneer might send a message to all the bidders that includes the following `received_orders` header:
+
+```json
+"received_orders": [
+  {"id": "did:ex:bob", "last": 2, "gaps": []},
+  {"id": "did:ex:alice", "last": 4, "gaps": [2,3]}      
+]
+```
+
+This lets Alice (or Bob) notice that the auctioneer hasn't seen the messages that Alice numbered 2 and 3. Gaps can then be plugged or ignored, depending on protocol rules. 


extensions/advanced_sequencing/main.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Scope
+
+This extension does not specify how gaps or out-of-order messages should be *handled* -- only how they should be *detected*. That is because some protocols may consider out-of-order messages harmless, whereas other protocols may be highly sensitive to problems triggered by such conditions. Thus, the scope of this extension is *detecting sequencing issues*. Any protocol that cares about sequencing issues SHOULD use this mechanism as its foundational semantics for understanding the problem, and MAY choose to react to the issues in whatever way the protocol designers deem appropriate. 
+
+## Reference
+
+DIDComm messages that support this extension offer three headers related to sequencing: `sender_order`, `sent_count`, and `received_orders`. These headers are only relevant to certain protocols. Protocol definitions MUST declare whether they use any of these headers. This is because general implementations of DIDComm are not required to implement support for the headers unless/until they intend to support protocols that use them.
+
+### `sender_order`
```
kdenhartog, line 15: Regarding `sender_order` and `sent_count` in concept are interesting and behave similarly to a TTL for cellular networks. However, one of the interesting things to note is that the trust the cellular network places on their devices (which they don't control) have allowed for an interesting vulnerability described in this [video](https://www.youtube.com/watch?v=alMsu_tfZPE).

My take away from this is that while `sender_order` and `sent_count` are useful, they couldn't be trusted by the recipient. For example, Mallory could start by sending `sender_order=10` and `sent_count=4` in an effort to disrupt thing so we should make sure that the protocol designers are aware of this limitation.

extensions/advanced_sequencing/main.md
```diff
@@ -0,0 +1,34 @@
...
+
+### `sent_count`
+
+The `sent_count` header enables resend behavior. A resent message contains the same headers (including `id` and `sender_order`) and body as a previous transmission. The first time a message is sent, the `sent_count` for that message is 1, and the header is normally omitted (the value of the header is implicitly 1). The second time the message is sent, the `sent_count` is 2, and the header is added to indicate that the message might be redundant ("this is the second time I've sent this to you"). The header continues to be incremented with each subsequent resend. Once a recipient has received one copy of a message, they MUST ignore subsequent copies that arrive, such that resends are idempotent.
+
+### `received_orders`
```
kdenhartog, line 23: Thoughts on renaming to `order_received`?

extensions/advanced_sequencing/main.md
```diff
@@ -0,0 +1,34 @@
...
+
+The `received_orders` header tells other parties how the sender has experienced the unfolding interaction so far. This allows gaps to be detected, possibly triggering a resend. The value of this header is an array of *gap detector* objects in the format `{"id":<did of party>, "last":<value of biggest sender_order seen from this party>, "gaps": []}`. In our running auction example, if the auctioneer sees bids 1 and 2 from Bob, and bid 1 and 4 from Alice, the auctioneer might send a message to all the bidders that includes the following `received_orders` header:
+
+```json
+"received_orders": [
```
kdenhartog, line 28: I'm concerned that this list could make messages rather large in unnecessary ways. Imagine the case where an exchange is managing an orderbook for millions of traders. My phone probably couldn't process the entire orderbook in the same way an exchanges servers could (nor would I want to consume that much bandwidth).

Could we add a method to limit the responses in some way while also allowing for the entire list to be sent in a debugging mode?

kdenhartog (commented):
In general, I think some functionality like this would be useful for protocol designers to optionally leverage. With a few additions I think this would be useful to keep.

extensions/advanced_sequencing/main.md
dhh1128: I agree that a malicious *sender* could send bad data. Nobody else could alter what the sender declares, however -- since these headers are in the plaintext that gets encrypted.

I agree that we should write up a security note about this. However, this vulnerability is not really very specific to this extension; it exists with the send time in core headers as well.

extensions/advanced_sequencing/main.md
dhh1128: I could go for `orders_received`. I want it to be plural, since it's an array.

extensions/advanced_sequencing/main.md
dhh1128: We could add a sentence saying that the gaps don't have to be complete, and give your example as a scenario where a tiny subset of what's possible would instead be recommended?

extensions/advanced_sequencing/main.md
kdenhartog: +1 to security note for this. In most cases, this is a useful feature and really all we're trying to achieve is make note that the protocol designers need to make note of these things in their security considerations section.

extensions/advanced_sequencing/main.md
kdenhartog: Yeah I agree. This sounds like details left to the implementers guide too now that I think about it since it's more implementation specific in how they handle it.

extensions/advanced_sequencing/main.md
```diff
@@ -0,0 +1,34 @@
...
+
+### `sender_order`
+
+The `sender_order` header tells other parties how the sender of a message perceives their own ordering. Its value is a monotonically increasing natural/ordinal number (1..<var>N</var>) that tells how many different messages the sender has sent in the current thread. When Alice and Bob are both bidding in an auction protocol, each of them marks their first bid with `sender_order: 1`, their second bid with `sender_order: 2`, and so forth. This allows the auctioneer to detect if Alice's bids arrive in a different order than she intended. It also means that any message can be uniquely identified by its `thid`, the sender, and the value of that sender's `sender_order`. Note how this does NOT clarify the sequence of Alice's messages relative to Bob's.
```
kdenhartog, line 17:
```suggestion
The `sender_order` header tells other parties how the sender of a message perceives their own ordering. Its value is a monotonically increasing natural/ordinal number (1..<var>N</var>) that tells how many different messages the sender has sent in the current thread. When Alice and Bob are both bidding in an auction protocol, each of them marks their first bid with `sender_order: 1`, their second bid with `sender_order: 2`, and so forth. This allows the auctioneer to detect if Alice's bids arrive in a different order than she intended. It also means that any message can be uniquely identified by its `thid`, the sender, and the value of that sender's `sender_order`. Note how this does NOT clarify the sequence of Alice's messages relative to Bob's. Due to the way `sender_order` relies upon the sender being non-malicious, protocol designers SHOULD make mention of the impact of this feature within their security considerations section.
```

extensions/advanced_sequencing/main.md
```diff
@@ -0,0 +1,34 @@
...
+
+### `sent_count`
+
+The `sent_count` header enables resend behavior. A resent message contains the same headers (including `id` and `sender_order`) and body as a previous transmission. The first time a message is sent, the `sent_count` for that message is 1, and the header is normally omitted (the value of the header is implicitly 1). The second time the message is sent, the `sent_count` is 2, and the header is added to indicate that the message might be redundant ("this is the second time I've sent this to you"). The header continues to be incremented with each subsequent resend. Once a recipient has received one copy of a message, they MUST ignore subsequent copies that arrive, such that resends are idempotent.
```
kdenhartog, line 21:
```suggestion
The `sent_count` header enables resend behavior. A resent message contains the same headers (including `id` and `sender_order`) and body as a previous transmission. The first time a message is sent, the `sent_count` for that message is 1, and the header is normally omitted (the value of the header is implicitly 1). The second time the message is sent, the `sent_count` is 2, and the header is added to indicate that the message might be redundant ("this is the second time I've sent this to you"). The header continues to be incremented with each subsequent resend. Once a recipient has received one copy of a message, they MUST ignore subsequent copies that arrive, such that resends are unchanged. Due to the way `sender_order` relies upon the sender being non-malicious, protocol designers SHOULD make mention of the impact of this feature within their security considerations section.
```

kdenhartog (approved):
Pending the suggested text being acceptable with @dhh1128 I think this looks ready to go. LGTM

PR #211, opened 2021-06-22
Attachment Format Attribute

TelegramSam:


kdenhartog (commented):
Is this mainly for instances where things get base64 encoded and stuff like that?

TelegramSam:
The main reason was knowing what the contents are besides json, etc. json, but which json? In the Aries Cred exchange protocols, there was an additional attribute used to indicate the format of the credential outside of the attachment. This is an attempt to make this more clear.

kdenhartog (approved):
Makes sense to me now. Thanks for clarifying it's purpose.

tplooker:
Discussed on 28th June WG call resolved to merge

PR #212, opened 2021-06-23
refactor: updated encryption spec based on ECDH-1PU draft4 update

### LINKED ISSUE #205: set skid/combined recipients kid values in APU/APV respectively as protected headers
The 1PU draft4 [example B](https://github.com/NeilMadden/jose-ecdh-1pu/blob/master/draft-madden-jose-ecdh-1pu-04/draft-madden-jose-ecdh-1pu-04.txt#L794) has the recipient headers stored in the ProtectedHeaders, particularly the APU and APV fields representing the sender (producer) and recipients (combined receivers), in our Aries RFC 334, we proposed to add the recipients’ kids in the AAD section (see authcrypt example [here](https://github.com/hyperledger/aries-rfcs/tree/master/features/0334-jwe-envelope#authcrypt-using-ecdh-1pu-key-wrapping-mode)).

I propose we set the the sender kid in APU (same value as skid) and the list of recipient’s kids in APV since draft4 is using the common protectedHeaders for these. The recipient kid remains under the recipients headers as per the example, only the other headers are now common for all recipients since we use a common epk for all recipients and the content encryption is done prior to key wrapping (since the encryption tag is used in the key wrapping process). This means we don’t need to set the list of recipients in AAD anymore.

Baha-sk:
closes #205

Signed-off-by: Baha Shaaban <baha.shaaban@securekey.com>

### DIFF
--- a/docs/spec-files/encryption.md
+++ b/docs/spec-files/encryption.md
@@ -26,10 +26,11 @@ For the keys involved in key agreement, the following elliptic curves MUST be su
 
 For content encryption of the message, the following algorithms MUST be supported.
 
-| Algorithm(JWA) | Description                |
-| -------------- | -------------------------- |
-| XC20P          | XChaCha20Poly1305          |
-| A256GCM        | AES-GCM with a 256 bit key |
+| Algorithm(JWA) | Description                | Authcrypt/Anoncrypt            |
+| -------------- | -------------------------- |------------------------------- |
+| XC20P          | XChaCha20Poly1305 with a 256 bit key | Anoncrypt |
+| A256GCM        | AES256-GCM with a 256 bit key | Anoncrypt |
+| A256CBC-HS512  | AES256-CBC + HMAC-SHA512 with a 512 bit key | Authcrypt/Anoncrypt |
 
 TODO: Include language about safe nonce considerations.
 
@@ -118,6 +119,21 @@ When the `skid` cannot be revealed in a plain-text JWE header (to avoid potentia
 For applications that don't require this protection, they MAY use `skid` protected header directly without any additional recipient headers.
 
 Applications MUST use either `skid` protected header or `encrypted_skid` recipients header but not both in the same envelope.
+
+## ECDH-1PU key wrapping and common protected headers
+When using authcrypt, the 1PU draft [requires](https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#section-2.1) mandates the use of AES_CBC_HMAC_SHA family of content encryption algorithms. To meet this requirement, JWE messages MUST use common `epk`, `apu`, `apv` and `alg` headers for all recipients. They MUST be set in the `protected` headers JWE section.
+
+As per this requirement, the JWE building must first encrypt the payload then use the resulting `tag` as part of the key derivation process when wrapping the `cek`.
+
+To meet this requirement, the above headers must be defined as follows:
+* `epk`: generated once for all recipients. It MUST be of the same type and curve as all recipient keys since kdf with the sender key must be on the same curve.
+  - Example: `"epk": {"kty": "EC","crv": "P-256","x": "BVDo69QfyXAdl6fbK6-QBYIsxv0CsNMtuDDVpMKgDYs","y": "G6bdoO2xblPHrKsAhef1dumrc0sChwyg7yTtTcfygHA"}`
+* `apu`: similar to `skid`, this is the producer (sender) identifier, it MUST contain the `skid` value base64 RawURL (no padding) encoded. Note: this is base64URL(`skid` value).
+  - Example for `skid` mentioned in an earlier [section](#key-ids-kid-and-skid-headers-references-in-the-did-document) above: `ZGlkOmV4YW1wbGU6MTIzNDU2Nzg5YWJjZGVmZ2hpI2tleXMtMQ`
+* `apv`: this represents the recipients' `kid` list. The list must be alphanumerically sorted, `kid` values will then be concatenated with a `.` and the final result MUST be base64 URL (no padding) encoding of the SHA256 hash of concatenated list.
+* `alg`: this is the key wrapping algorithm, ie: `ECDH-1PU+A256KW`.
+
+A final note about `skid` header: since the 1PU draft [does not require](https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#section-2.2.1) this header, authcrypt implementations MUST be able to resolve the sender kid from the `APU` header if `skid` is not set.
 
 ## Examples
 


kdenhartog (changes requested):
I appreciate the effort to align everything. That's definitely not something easy to do. With that in mind, we intentionally discussed limiting the algorithms supported in this WG and landed on these choices for various reasons.

I'd not be in favor of introducing P-521 at this point since I think it introduces more variance than necessary. In fact, I'd prefer to drop any mention of P-256 even.

On the topic of AES CBC, I think we need to work with @NeilMadden to figure out what's driving the requirement for CBC. As far as I can tell, A256GCM or XC20P are both viable options for use with 1PU and we may be being overly restrictive here.

It should be noted that from my read of the paper, message franking is a way to introduce non-repudiation in symmetrical authenticated encryption schemes, so we may be faced with additional tradeoffs that have to be considered here.

Can we hold off on merging this until we've had a proper ability to evaluate the variety of considerations?

kdenhartog:
Did a further review of the [paper](https://link.springer.com/content/pdf/10.1007%2F978-3-319-63697-9_3.pdf). Looks like the reason the AES-GCM and XChacha20Poly1305 doesn't work comes down to the way in which the MACs are built doesn't allow for recipient binding. I don't fully comprehend the how/why part yet, but figured I'd weigh in on that finding now.

Still, I think we need to get some expertise on the tradeoffs here because it definitely seems like we're going to encounter them here. My suggestion is we first figure out what the tradeoffs are (e.g. non-repudiation by default is a no so no message franking) and then figure out what crypto meets our requirements.

NeilMadden:
The reason for requiring CBC+HMAC encryption is given in paragraph 5 of https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#section-4 

Essentially, when sending a message to multiple recipients with key-wrapping mode, a single Content Encryption Key (CEK) is used to encrypt the message. This means that any one of those recipients can reuse that CEK to create a new message and send it to the original recipients - who will all think it comes from the sender of the original message. This is sometimes referred to as a lack of *insider security*. ECDH-1PU draft 4 fixes this by making the key derivation process depend on the entire contents of the message - so the CEK is effectively bound to that single message and can't be re-used. It does that efficiently by simply including the MAC tag from the AEAD computation in the KDF. But this is only secure if the MAC tag is compactly committing - i.e. someone who knows the CEK cannot find another message with the same tag (this is possible for GCM and ChaPoly).

Regarding non-repudiation, the solution given here doesn't perform message franking, it just reuses some security notions from that paper. You could describe ECDH-1PU draft 4 as providing "insider non-repudiation": if Alice sends a message to Bob and Charlie, then Alice, Bob, and Charlie can all prove to each other that Alice is the originator of that message, but they can't prove it to anyone else who wasn't in the recipient list. IMO this is the security goal you usually want.

kdenhartog:
Awesome thanks for weighing in on this. 

> But this is only secure if the MAC tag is compactly committing - i.e. someone who knows the CEK cannot find another message with the same tag (this is possible for GCM and ChaPoly).

I was trying to understand this part of the proof for my own comprehension, so thanks for putting it in layman's terms.

>Regarding non-repudiation, the solution given here doesn't perform message franking, it just reuses some security notions from that paper. You could describe ECDH-1PU draft 4 as providing "insider non-repudiation": if Alice sends a message to Bob and Charlie, then Alice, Bob, and Charlie can all prove to each other that Alice is the originator of that message, but they can't prove it to anyone else who wasn't in the recipient list. IMO this is the security goal you usually want.

Yeah this does sound like what we're after. I didn't catch that nuance about reusing a security property described originally in that paper.

Ok, I'm convinced that AES-CBC is probably the way to go for us. Thanks for your input on this @NeilMadden

On the bright side, it should help us consolidate a bit further on the cryptographic primitives in use.

troyronda:
We still have P256 listed in the table, so I think it should be P256/P384/X25519 for now.

@Baha-sk @kdenhartog

kdenhartog:
Yeah, that's what I had remembered as well. I just figured because we're still in draft stage that we'd eventually drop it before we go to 1.0. This is similar to how did-core treated `publicKey` when replacing it with `verificationMethod`

Baha-sk:
Thank you @NeilMadden @kdenhartog and @troyronda for your valuable comments.

For Anoncrypt, the sender identity is not revealed since we will be using ECDH-ES with KW of the cek. We should follow the JWA spec. As per [section 5.1](https://datatracker.ietf.org/doc/html/rfc7518#section-5.1), these are the 2 mandatory content encryptions: `A128CBC-HS256` and `A256CBC-HS512`, all other algs are optional/recommended. I will keep Anoncrypt permissive (all AES-GCM and AES-CBC algs are optional except for the two above that are mandatory).

Also since there's interest in XC20P, which is not standard in JWE, it will be optional for Anoncrypt only.

For Authcrypt though, only AES-CBC algs must be supported in light of the previous comments above and since they're mandated by the 1PU draft.

Finally as for the curves supported in recipient/sender keys, we should be permissive to support all major curves (P-256, P-384, P-521 and X25519) (no RSA). We can keep P-256 deprecated but I think P-521 should be on the list as well as in this pr.

Baha-sk:
I've updated the PR to only include A256CBC-HS512 as the new content algorithm required following 1PU draft4 update. Other CBC algs are optional.

Same for the EC curves, I've kept mandatory curves only.

kdenhartog (approved):
Looks good to me after the discussion here.

docs/spec-files/encryption.md
```diff
@@ -26,10 +26,11 @@ For the keys involved in key agreement, the following elliptic curves MUST be su
...
 
-| Algorithm(JWA) | Description                |
-| -------------- | -------------------------- |
-| XC20P          | XChaCha20Poly1305          |
-| A256GCM        | AES-GCM with a 256 bit key |
+| Algorithm(JWA) | Description                | Authcrypt/Anoncrypt            |
+| -------------- | -------------------------- |------------------------------- |
+| XC20P          | XChaCha20Poly1305 with a 256 bit key | Anoncrypt |
+| A256GCM        | AES256-GCM with a 256 bit key | Anoncrypt |
+| A256CBC-HS512  | AES256-CBC + HMAC-SHA512 with a 512 bit key | Authcrypt/Anoncrypt |
```
troyronda, line 33: I think you meant a 256 bit key. @Baha-sk

awoie:
@NeilMadden would it be possible to support XC20PKW for key wrapping + XC20P for content encryption (as defined by  [draft-amringer-jose-chacha-02](https://datatracker.ietf.org/doc/html/draft-amringer-jose-chacha-02)) additionally by the ECDH-1PU spec?

NeilMadden:
As mentioned above, (X)ChaCha20-Poly1305 is not suitable for use with ECDH-1PU as it is not compactly committing and so leads to security vulnerabilities in the multi-recipient case. With regards to the key-wrap variant, ChaPoly does not meet the usual [security definitions of a KeyWrap algorithm](https://www.cs.ucdavis.edu/~rogaway/papers/keywrap.pdf). It seems similar to the AES-GCM key-wrap algorithm defined in JWA, which I don't think I've ever seen used in reality and it's not at all clear why it was invented. (I'd much rather see a proper [SIV mode](https://datatracker.ietf.org/doc/html/draft-madden-jose-siv-mode-02) for JOSE). That said, so long as you use a fresh IV then it is presumably secure - it just seems quite an inefficient construction for the job.

awoie:
> As mentioned above, (X)ChaCha20-Poly1305 is not suitable for use with ECDH-1PU as it is not compactly committing and so leads to security vulnerabilities in the multi-recipient case. With regards to the key-wrap variant, ChaPoly does not meet the usual [security definitions of a KeyWrap algorithm](https://www.cs.ucdavis.edu/~rogaway/papers/keywrap.pdf). It seems similar to the AES-GCM key-wrap algorithm defined in JWA, which I don't think I've ever seen used in reality and it's not at all clear why it was invented. (I'd much rather see a proper [SIV mode](https://datatracker.ietf.org/doc/html/draft-madden-jose-siv-mode-02) for JOSE). That said, so long as you use a fresh IV then it is presumably secure - it just seems quite an inefficient construction for the job.

Thanks for the reply. That clarifies my question.

awoie:
PR lgtm. IMO, we could even drop XC20P for content encryption.

Baha-sk:
@NeilMadden when I was implementin 1PU in Go, I've asked Google Tink folks if they do support AES-CBC+HMAC-SHA (because we use Tink to manage keys and execute crypto primitives), the answer I got was that AES-CBC is inferior to AES-CTR mode and they only support AES-CTR_HMAC-SHA.

I had to implement a Tink key that support AES-CBC+HMAC-SHA [internally in  go](https://github.com/hyperledger/aries-framework-go/tree/main/pkg/crypto/tinkcrypto/primitive/aead) to be compliant with the 1PU draft.

My question to you: is it really better to use AES-CTR mode instead of AES-CBC?

I noticed that the JWE [approved algorithms](https://datatracker.ietf.org/doc/html/rfc7518#section-5.1) don't have CTR mode in the list. Will this be a challenge to add if really AES-CTR is better than AES-CBC? Do you have an idea why it's not on the list?

Thanks.

NeilMadden:
Yes, the modern view is that CTR-mode is [the best confidentiality-only block cipher mode](https://web.cs.ucdavis.edu/~rogaway/papers/modes.pdf). I don't know why JOSE specified CBC+HMAC and not CTR+HMAC, except that CBC is perhaps a more widely implemented mode historically (e.g., Signal uses CBC+HMAC, as does Fernet). In the specific context of JOSE, the benefits of CTR over CBC are largely academic.

Baha-sk:
Thank you very much @NeilMadden  for the explanation, I've got a similar answer from @kdenhartog yesterday as well. 

To remain compliant with JWE, we'll stick with `A256CBC-HS512` as the default required algorithm.

PR #227, opened 2021-07-12
Adjust language around sender key protection

andrewwhitehead:
A first draft, as discussed in the June 14th DIDComm WG meeting


dhh1128 (commented):
I find myself wanting to add a sentence that explains that at every level of wrapping, the sender's DID and key can be different. Outer envelopes can use anoncrypt (in which case the DID of the sender will mean nothing to the mediator or parties that can observe the plaintext seen by the mediator). It's mainly the innermost envelope that matters here -- so the only eavesdropper that degrades the story here is the eavesdropper on an agent that terminates a delivery route. Right?

Issue #226, opened 2021-07-10
Are we going to address Internationalization / localization?


TelegramSam:
DIDComm V1 has a way to indicate language and alternative translations within a message. Is that something we wish to include in v2?

kdenhartog:
We'll need to decide if we want to defer this issue or set it as a v2 issue as well. I tend to think we're going to need to address this in a baseline V2 although I have no expertise in this area.

kdenhartog:
@dhh1128 said on #152
> We need enough i18n support in DIDComm v2 to prevent a variety of incompatible solutions from arising. We should explain how a DIDComm protocol can solve 2 problems:
> 
>     1. Clarify the lang of a field ("These instructions for the loan applicant are in Spanish.")
> 
>     2. Allow fields in a message to have the same semantic role, but provide content in different languages. ("Here's the French equivalent of the Spanish instructions field.")
> 
> 
> This is likely to require much less complexity than the i18n mechanism proposed for DIDComm v1 -- but perhaps it will use a subset of it?

Moving the comment here so that we can consolidate the discussions

TelegramSam:
closed, given the approved content.

PR #235, opened 2021-07-19
First cut at i18 section

dhh1128:

This content is not ready to merge, because the new file isn't referenced in the spec.js file anywhere. I wanted to see if we agreed on the ideas before inserting them somewhere.


dhh1128:
Fixes #226

dhh1128:
Remove inline.
Define error code for "Can't read that language. Can you send me this other one? `<lang code>`"
Define how to use Feature Discovery to find out the lang of the other party.
Make `lang` setting sticky.

kdenhartog (approved):
I'm liking where this is going

TelegramSam:
I like this. I have a minor preference to introduce the `lang` header first, and then the `accept-lang`, but  that is a super minor point.

I love that you included the inline mode as an extension. This preserves the thinking and work, but does not require it, lowering adoption hurdles.

I think this is ready to merge.

awoie:
As per WG meeting from August 2nd, merging this PR

PR #240, opened 2021-07-24
Clarify some minor details about attachments.

dhh1128:


dhh1128:
Most of the changes here are simple and uncontroversial; they simply clarify something that I believe is already consensus tribal knowledge. However, there is one change that should be discussed: I am proposing that we remove `jwe`. The reason is that `jwe` is redundant. We can already attach a JWE to a message just by putting it in the `data.base64` field. (Are we using the `jwe` field to optimize away the base64-re-encoding a JWM from a deeper layer of the routing "onion"? If so, we could do that in the `json` field; we still don't need `jwe` for it.)

We could say that we have `jws`, so we need `jwe`. But this would be a false parallel. The `jws` field is in detached mode (is it not?). So it doesn't provide data; it provides only the signature over the data.

docs/spec-files/general_messaging_constructs.md
```diff
@@ -24,19 +24,18 @@ Attachments are contained within a list in the `attachments` header.
...
+- `id`: [optional but recommended] Identifies attached content within the scope of a given message, so it can be referenced. For example, in a message documenting items for sale on an auction website, there might be a field named `front_view` that contains the value `#attachment1`; this would reference an attachment to the message with `id` equal to `attachment1`. Recommended on appended attachment descriptors. If omitted, then there is no way to refer to the attachment later in the thread, in error messages, and so forth. Because `id` is used to compose URIs, this value should be brief and MUST consist entirely of [unreserved URI characters](https://datatracker.ietf.org/doc/html/rfc3986/#section-2.3) – meaning that it is not necessary to [percent encode](https://en.wikipedia.org/wiki/Percent-encoding) the value to incorporate it in a URI.
```
kdenhartog, line 27:
```suggestion
- `id`: [recommended] Identifies attached content within the scope of a given message, so it can be referenced. For example, in a message documenting items for sale on an auction website, there might be a field named `front_view` that contains the value `#attachment1`; this would reference an attachment to the message with `id` equal to `attachment1`. Recommended on appended attachment descriptors. If omitted, then there is no way to refer to the attachment later in the thread, in error messages, and so forth. Because `id` is used to compose URIs, this value should be brief and MUST consist entirely of [unreserved URI characters](https://datatracker.ietf.org/doc/html/rfc3986/#section-2.3) – meaning that it is not necessary to [percent encode](https://en.wikipedia.org/wiki/Percent-encoding) the value to incorporate it in a URI.
```

Would it make sense to use urn:didcommv2:id:<message-thread-id>#<attachmentId> so that we can reference attachments across messages? Id prefer to default to using just #<attachmentId> for simplicity, but thinking this would be beneficial in some optimization cases.

docs/spec-files/general_messaging_constructs.md
```diff
@@ -24,19 +24,18 @@ Attachments are contained within a list in the `attachments` header.
...
-- `data`: A JSON object that gives access to the actual content of the attachment. Contains the following subfields:
-  - `jws`: [optional] A JSON Web Signature over the content of the attachment.
+- `data`: A JSON object that gives access to the actual content of the attachment. Contains enough of the following subfields to allow access to the data:
+  - `jws`: [optional] A [JWS](https://tools.ietf.org/html/rfc7515) in [detached dontent mode](https://tools.ietf.org/html/rfc7515#appendix-F), where the `payload` field of the JWS maps to `base64` or to something fetchable via `links`. This allows attachments to be signed. The signature need not come from the author of the message.
   - `hash`: [optional] The hash of the content encoded in multi-hash format. Used as an integrity check for the attachment, and MUST be used if the data is referenced via the `links` data attribute. 
   - `links`: [optional] A list of zero or more locations at which the content may be fetched.
-  - `base64`: [optional]Base64-encoded data, when representing arbitrary content inline instead of via `links`.
-  - `jwe`: Already encrypted data in the form of a JWE.
```
kdenhartog, line 39: Would we need this for when we're sending a JWE JSON message?

docs/spec-files/general_messaging_constructs.md
brianorwhatever: yes always fully qualified

docs/spec-files/general_messaging_constructs.md
brianorwhatever: ie. forward protocol I believe uses this?

it is just json though.. which begs the question why have all of these keys at all. why not just have 3? `id`, `type`, `base64`

docs/spec-files/general_messaging_constructs.md
dhh1128: There is a whole proposal about DIDComm message hyperlinking. I tried to put it in the DIDComm spec but was told to take it out as unnecessary. But the requirements as written are intended to support linking across messages. You don't need to make the references fully qualified *inside* a message (this isn't JSON-LD where we are trying to canonicalize and sign); you just need to know how to convert a relative reference to absolute.

docs/spec-files/general_messaging_constructs.md
dhh1128: I actually proposed just `base64`, but was outvoted. I still like that.

I *could* make one argument for `json`: it prevents an extra 33% bloat. If I put a JWE into `base64`, I have to base64 a bunch of stuff that's already base64'ed internally, so the bloat is worse. Having `json` lets me skip this.

But I can't really see *any* justification for both `json` and `jwe`. JWE is just JSON -- and we can already tell them apart using `media-type`.

docs/spec-files/general_messaging_constructs.md
brianorwhatever: I prefer not to do the conversion so as long as I can use fully qualified which I assume is still valid I am fine with relative.

docs/spec-files/general_messaging_constructs.md
kdenhartog: I too wonder what the need for the property to describe it inside data is for. Now that we have the mime-type I wonder if it's useful for us to just put all the data as the value in the "data" property and make the value polymorphic (e.g. could be a JSON array, string, number, object)

TelegramSam: @kdenhartog I'd expect pushback on having a value be of a variable type for those using inflexible serialization engines.

TelegramSam:
I like these simplificaitons.

TelegramSam (approved):

awoie:
As per WG meeting from August 2nd, merging this PR.

PR #244, opened 2021-07-30
CBOR commentary

### LINKED ISSUE #207: Should we flip to CBOR?
Before we ever began the DIDComm v2 effort, I had a conversation with @tplooker about the advisability of basing DIDComm on something binary instead of on JSON. I told him then that I thought we could go full-on binary if we did it from the inception of the v2 effort, but that changing to binary once we were well underway would be too painful. So my advice then was to either propose CBOR immediately, or leave it alone.

(History: the *very first* ancestor of modern DIDComm, called A2A in the what was then the Hyperledger Indy community, was implemented with binary serialization using MsgPack. We had an IIW meeting soon after that where we talked about message formats, and I came to a session called by Sam C where the question of these formats was debated a bit. I was passionately in favor of binary over JSON because we already had something working, and I didn't want to have to change it. But I got voted down, and we became very JSON-oriented for all of DIDComm v1 -- and then, by inheritance, DIDComm v2.)

I've been content to let this decision ride, but today I had an interesting conversation with some advocates of KERI that left me feeling really discouraged about the semantic and architectural overhead we're incurring, constantly, by being JSON-centric. (This isn't a comment about KERI, particularly -- and the KERI proponents I was talking to weren't dissing JSON, they were defending it. But their defense left me squirming. The impedance mismatch between JSON and communication features rears its head periodically in lots of contexts. It just happened to be a KERI conversation where my discomfort came to a head today.)

Now, I know that it is very late in the DIDcomm v2 lifecycle to be asking this question, and that I am contradicting my own earlier advice to even ask the question again. But what I'm proposing to do in this issue is to document for posterity (in this issue's comment stream) why we chose to base DIDComm v2 on JSON instead of CBOR. Then, if we are content with the outcome, I'll be content as well -- and we'll lay this issue to rest forever, with a citable public conversation about why. If, on the other hand, others begin to squirm in the same way I did, we could debate whether we want to change our minds.

TelegramSam:
Fixes #207 

### DIFF
--- a/docs/spec-files/future_proofing.md
+++ b/docs/spec-files/future_proofing.md
@@ -18,6 +18,8 @@ Each DIDComm extension is described in a specification of its own. Software that
 
 DIDComm Messages are JSON encoded (based on the JOSE family of specs) at the encryption, signature, and content level. Future encodings might introduce binary serializations. Each innovation like this MUST specify a deterministic and reliable method for indicating the alternative encoding used.
 
+At multiple points in the creation of this spec the community discussed switching to CBOR as a primary encoding format to replace JSON. Two reasons prevented that switch: The maturity of the JSON related standards as compared to CBOR related standards, and the cost of switching so late in spec development.
+
 ### Beyond Messaging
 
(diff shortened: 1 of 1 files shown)


kdenhartog (approved):

dhh1128 (approved):

awoie:
As per WG meeting from August 2nd, merging this PR

PR #246, opened 2021-08-03
Clarify that rewrapping means re-encrypting

dhh1128:

Answers a question raised by implementers on the Gemini project: https://didcomm-gemini.slack.com/archives/C027QMXHWF8/p1628010475070000

### DIFF
--- a/docs/spec-files/routing.md
+++ b/docs/spec-files/routing.md
@@ -68,7 +68,7 @@ The attachment(s) in the `attachments` field are able to use the full power of D
 
 #### Rewrapping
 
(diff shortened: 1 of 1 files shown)


dhh1128:
The alternative view of rewrapping is that the new `forward` message is left as plaintext. This does not destroy all security guarantees, since the payload of the new `forward` message remains encrypted. However, it means that intermediate parties that participate in the rewrapping all get to see the `skid` of the innermost message, which lessens privacy to some degree. Also, eavesdroppers learn that the messages being sent to parties participating in the rewrapping scheme are `forward` messages instead of just opaque DIDComm messages. That is why I was advocating that rewrapping = reencryption.

dhh1128:
Tagging @ashcherbakov

TelegramSam (approved):
I'm a fan of this clarification

kdenhartog (approved):

PR #258, opened 2021-08-23
Some editorial changes

awoie:
Some editorial changes

### DIFF
--- a/docs/spec-files/processing_model.md
+++ b/docs/spec-files/processing_model.md
@@ -1,5 +1,5 @@
 ### Processing Model
 
 - Message-oriented, with sender and receiver(s)
-- Encryption Envelope
+- Encryption envelope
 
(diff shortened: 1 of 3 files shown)


awoie:
I'm intending to do some editorial PRs to clean up the spec a bit, e.g., fixing links, highlighting, formatting etc.

kdenhartog (approved):
I personally am not a fan of personification in specs because I believe it often leads to overly constraining the mental model. In this case, I don't think it's doing that and I recognize that this is consistent with other parts of the spec so still going to approve it.

awoie:
> I personally am not a fan of personification in specs because I believe it often leads to overly constraining the mental model. In this case, I don't think it's doing that and I recognize that this is consistent with other parts of the spec so still going to approve it.

Ah, you are referring to the Alice and Bob examples. Yes, the intention is that we should harmonize language. In some sections we use Alice, Bob, Carol, in other parts we say Agent A, B etc.

TelegramSam (approved):

PR #268, opened 2021-08-30
potentially fixes #257; made serviceEndpoint an URI or object

### LINKED ISSUE #257: Allow for objects as serviceEndpoints
According to DID Core, objects (set/map) are possible too. The current DIDComm v2 spec only allows URIs for service endpoints but we should rather let the specific transport decide what form the `serviceEndpoint` should have. 

Background: we are working on a transport that will likely require an object rather than a URI.

awoie:
potentially fixes #257; made serviceEndpoint an URI or object

### DIFF
--- a/docs/spec-files/routing.md
+++ b/docs/spec-files/routing.md
@@ -129,7 +129,7 @@ DIDComm DID Document endpoints have the following format:
 
 **type**: MUST be `DIDCommMessaging`. 
 
-**serviceEndpoint**: MUST contain a URI for a transport specified in the [transports] section of this spec, or a URI from Alternative Endpoints. It MAY be desirable to constraint endpoints from the [transports] section so that they are used only for the reception of DIDComm messages. This can be particularly helpful in cases where auto-detecting message types is inefficient or undesirable.
+**serviceEndpoint**: MUST be an URI or object for a transport specified in the [transports] section of this spec, or an URI or object from alternative endpoints. It MAY be desirable to constraint endpoints from the [transports] section so that they are used only for the reception of DIDComm messages. This can be particularly helpful in cases where auto-detecting message types is inefficient or undesirable.
 
 **accept***: An optional array of media types in the order of preference for sending a message to the endpoint.
 If `accept` is not specified, the sender uses its preferred choice for sending a message to the endpoint.


kdenhartog (approved):

awoie:
Before we can merge, we should change the language. The WG was fine with the changes in general but requested to use the same language as in DID Core. Furthermore, in case the serviceEndpoint is a map, we should require an additional `transportType` property (URI) to indicate the actual transport. Note, for URIs, the type of transport can be usually derived from the URI scheme.

TelegramSam:
discussed changes that point to the DID Core spec.

TelegramSam:
Discussed WG 20220314. 

Discussed ION, which impl cannot accept the args as we've designed.
Suggestion to test the proposed object format to make sure it works.
related: https://github.com/decentralized-identity/didcomm-messaging/issues/294

general consensus was that making the adjustment might be easier than getting ION to fully support the did core spec as written. ION support is desired, either by contacting their team or by making the object adjustment.

We support moving AWAY from a string and to the full object support to avoid having two forms to parse. Need a PR with all of those changes to continue the discussion.

jayv80:
@TelegramSam I've checked today with my team and the Hedera identity fabric that we currently manage. We have the same problem as ION at the moment. 
We can only extend the ServiceEndpoint with the current implementation of the DID-Core spec. Meaning the ServiceEndpoint can be an object that contains specific attributes. 

This design choice was made due to the fact that is more logical to include specific attributes in the ServiceEndpoint ojbect, than sibling attributes in the Services property. 
Because, it could happen that during further extensions of the DID-Core specification, sibling attributes are in collision with each other.  

So this problem is not limited to ION alone. I honestly thing that we're not changing the current DID-Core implementation for Hedera in the short future.

TelegramSam:
> Because, it could happen that during further extensions of the DID-Core specification, sibling attributes are in collision with each other.

Given how few service endpoints there are, and DIDComm being a majority force I think it's unlikely that they would choose a colliding attribute. That said, knowing that you've made the same choice as ION is good input for our efforts.

TelegramSam:
Now irrellevant, fixed in PR #352

PR #267, opened 2021-08-27
feat: add ECDH-1PU article

baha-ai:
Signed-off-by: Baha Shaaban <baha.shaaban@securekey.com>

PR #270, opened 2021-08-30
update the number of recipients keys in a JWE envelope

Baha-sk:
Signed-off-by: Baha Shaaban <baha.shaaban@securekey.com>

### DIFF
--- a/docs/spec-files/encryption.md
+++ b/docs/spec-files/encryption.md
@@ -101,7 +101,7 @@ Assuming she also has Bob's DID document which happens to include the following 
 }
 ```
 
-There should be only 1 entry in the recipients of the envelope, representing Bob. The corresponding `kid` header for this recipient MUST have `did:example:jklmnopqrstuvwxyz1#key-1` as value. The packing process MUST extract the public key bytes found in `publicKeyBase58` of Bob's DID Doc `KeyAgreement[0]` to execute the ECDH-1PU key derivation for content key wrapping.
+Unless previously coordinated in a layer above DIDComm, the default recipients of the envelope MUST include all the `KeyAgreement` entries, representing Bob. The corresponding `kid` header for this recipient MUST have a DID URL pointing to a corresponding verification method in the DID document. This verification method MUST be associated with the `KeyAgreement` verification relationship and the verification material MUST be retrieved from the DID document to execute the ECDH-1PU key derivation for content key wrapping.
 
 When Bob receives the envelope, the unpacking process on his end MUST resolve the `skid` protected header value using Alice's DID doc's `KeyAgreement[0]` in order to extract her public key. In Alice's DID Doc example above, `KeyAgreement[0]` is a reference id, it MUST be resolved from the main `VerificationMethod[]` of Alice's DID document (not shown in the example).
 


docs/spec-files/encryption.md
```diff
@@ -101,7 +101,7 @@ Assuming she also has Bob's DID document which happens to include the following
 }
 ```
 
-There should be only 1 entry in the recipients of the envelope, representing Bob. The corresponding `kid` header for this recipient MUST have `did:example:jklmnopqrstuvwxyz1#key-1` as value. The packing process MUST extract the public key bytes found in `publicKeyBase58` of Bob's DID Doc `KeyAgreement[0]` to execute the ECDH-1PU key derivation for content key wrapping.
+Unless previously coordinated, the recipients of the envelope SHOULD include all the KeyAgreement entries, representing Bob. The corresponding `kid` header for this recipient MUST have a DID URL pointing to a verification method in the did document. This verification method MUST be associated with the `KeyAgreement` verification relationship and the public key bytes MUST be retrieved from the DID Document to execute the ECDH-1PU key derivation for content key wrapping.
```
kdenhartog, line 104:
```suggestion
Unless previously coordinated, the recipients of the envelope SHOULD include all the KeyAgreement entries, representing Bob. The corresponding `kid` header for this recipient MUST have a DID URL pointing to a verification method in the did document. This verification method MUST be associated with the `KeyAgreement` verification relationship and the verification material MUST be retrieved from the DID Document to execute the ECDH-1PU key derivation for content key wrapping.
```

kdenhartog (approved):
One nit that @awoie raised on the other PR. Otherwise looks good to go.

kdenhartog:
@dhh1128 I believe you had concerns with a MUST in this statement: "Unless previously coordinated, the recipients of the envelope SHOULD"

Thinking through this, I'm not exactly certain how the authorization would be abstracted to accept a key entry in the KeyAgreeement VM or some other arbitrary means of per-agreement. Could you clarify what it is that would be expected here to code this up? I'm concerned that we're leaving a potential hole in the authorization mechanism for binding the identifier (the DID) to the key material.

dhh1128:
@kdenhartog : I think the canonical behavior in "naked DIDComm" is to always encrypt for all keyAgreement keys of the recipient. However, I can imagine situations where a recipient and a sender have some higher-level conversation going on that tells the sender to send to less than all of the recipient's keys. As a simple example, imagine that the recipient is a company that has 100 agents, but the sender has been assigned to work with agent 27. In such a case, it might make sense for the sender to encrypt only for agent 27.

How this agreement between sender and recipient is consummated is out of scope of the main DIDComm spec, in my opinion. Perhaps it's information carried in a higher level protocol -- and it should stay outside the DIDComm spec. But I just wanted to weaking the original "MUST include all the keyAgreement entries" to a "SHOULD" to allow for the possibility that such a situation could happen. After discussing on the WG call, we decided that "SHOULD" needed to be paired with the earlier phrase, "Unless previously coordinated", to make sure that the freedom to do something like this is dependent on some other coordination, not on the whim of the sender. But with that caveat phrase, I would actually be okay with a return to "MUST": "Unless previously coordinated... the recipients MUST include."

Baha-sk:
> But with that caveat phrase, I would actually be okay with a return to "MUST": "Unless previously coordinated... the recipients MUST include."

I'll revert SHOULD to MUST in this case

kdenhartog:
> @kdenhartog : I think the canonical behavior in "naked DIDComm" is to always encrypt for all keyAgreement keys of the recipient. However, I can imagine situations where a recipient and a sender have some higher-level conversation going on that tells the sender to send to less than all of the recipient's keys. As a simple example, imagine that the recipient is a company that has 100 agents, but the sender has been assigned to work with agent 27. In such a case, it might make sense for the sender to encrypt only for agent 27.
> 
> How this agreement between sender and recipient is consummated is out of scope of the main DIDComm spec, in my opinion. Perhaps it's information carried in a higher level protocol -- and it should stay outside the DIDComm spec. But I just wanted to weaking the original "MUST include all the keyAgreement entries" to a "SHOULD" to allow for the possibility that such a situation could happen. After discussing on the WG call, we decided that "SHOULD" needed to be paired with the earlier phrase, "Unless previously coordinated", to make sure that the freedom to do something like this is dependent on some other coordination, not on the whim of the sender. But with that caveat phrase, I would actually be okay with a return to "MUST": "Unless previously coordinated... the recipients MUST include."

Thanks for the additional context, I'm good with `"Unless previously coordinated... the recipients MUST include."`

I think the "unless previously coordinated will be untestable at this point as well, so we're likely to have divergence in interop on that until we can formally define how that coordination would occur. That's alright in my opinion though because we're basically just introducing the possibility for extension here without needing to break compliance in the future.

Baha-sk:
> 
> I think the "unless previously coordinated" will be untestable at this point as well, so we're likely to have divergence in interop on that until we can formally define how that coordination would occur. That's alright in my opinion though because we're basically just introducing the possibility for extension here without needing to break compliance in the future. This is a perfect example of an "extension point" in my eyes.

Thanks @kdenhartog.

@awoie, @dhh1128 let me know if you're ok with this change as i moved it out of #267 (and clarified a bit what `previously coordinated` means).

kdenhartog (approved):

TelegramSam:
I'm also good with "Unless previously coordinated... the recipients MUST include."

awoie (approved):
lgtm - the encryption section will need some editorial changes later.

awoie:
As per 6 September WG meeting, merging this PR.

PR #272, opened 2021-08-31
Update trustping.md

### LINKED ISSUE #271: Trust Ping protocol needs a `from` attribute
In order to respond to a trust ping the receiver needs to know where to send the response. Is there a different way to determine the sender that I'm not aware of?

brianorwhatever:
closes #271

### DIFF
--- a/docs/spec-files/trustping.md
+++ b/docs/spec-files/trustping.md
@@ -55,8 +55,6 @@ creates a `ping` message like this:
 
 **response_requested**: default value is `true`. If false, the `sender` is not requesting a `ping_response` from the `receiver`. If `true`, the `sender` is requesting a response.
 
-**from**: is required if a response is requested so that the receiver knows where to send the response
-
 ##### ping_response
 
 When the message arrives at the receiver, assuming that `response_requested`
--- a/docs/spec-files/trustping.md
+++ b/docs/spec-files/trustping.md
@@ -46,6 +46,7 @@ creates a `ping` message like this:
 {
   "type": "https://didcomm.org/trust_ping/1.0/ping",
   "id": "518be002-de8e-456e-b3d5-8fe472477a86",
+  "from": "did:example:123456",
   "body": {
       "response_requested": true
   }
@@ -53,6 +54,8 @@ creates a `ping` message like this:
 ```
 
 **response_requested**: default value is `true`. If false, the `sender` is not requesting a `ping_response` from the `receiver`. If `true`, the `sender` is requesting a response.
+
+**from**: is required if a response is requested so that the receiver knows where to send the response
 
 ##### ping_response
 


TelegramSam:
I vote that `from` is required for this protocol, even if no response is required. The point of the ping is to verify a trusted relationship. If no `from`, no relationship.

TelegramSam:
> How is identifier authentication being done now? Couldn't we rely on that methodology to get it instead?

You mean authenticated encryption? That is the method I'm expecting we rely on.

brianorwhatever:
So I think the issue here is that the "Trust Ping" protocol becomes just the "Ping" protocol when completed using anonymous encryption. I don't believe there is a way to authenticate in this case and all we will be able to rely on is a self declared `from` value

swcurran:
If you just send out a trust ping to a previously unknown contact using anoncrypt, yes.  But you can use an existing connection and (I think, if I've been following the conversation), the message would be delivered using authcrypt.

I don't see a need to change the name.  The information about using the DIDComm 2 with and without a connection should be general information and need not be tied to each protocol.

TelegramSam:
I've proposed in #274 that we make from required. That clears up several things, including this confusion about a trustpring without a from. Thoughts?

docs/spec-files/trustping.md
```diff
@@ -54,6 +55,8 @@ creates a `ping` message like this:
 
 **response_requested**: default value is `true`. If false, the `sender` is not requesting a `ping_response` from the `receiver`. If `true`, the `sender` is requesting a response.
 
+**from**: is required if a response is requested so that the receiver knows where to send the response
+
```
awoie, line 59: @brianorwhatever I guess we can remove the `from` after we merged #274

awoie (changes requested):
@brianorwhatever I guess we can remove the from after we merged #274

docs/spec-files/trustping.md
awoie, line 59:
```suggestion
```

awoie (changes requested):
lgtm after removing the `from` language, but merge should wait until #274 got merged

docs/spec-files/trustping.md
brianorwhatever: thanks @awoie

kdenhartog:
> > 
> 
> You mean authenticated encryption? That is the method I'm expecting we rely on.

Yup, I've started to rephrase my language to prevent conflation between what we mean when we say "authenticated encryption"/Authcrypt and what's more commonly used in symmetrical encryption where they're describing a specific security property that's slightly different than what we mean.

In my mind identifier authentication utilizes traditional "authenticated encryption" (they normally just mean the encryption is integrity protected and tamper resistant by people without the secret key) and that we're binding the integrity to a specific identifier (e.g. the did) that can be inferred. This may not be exactly correct terminology, but just looking to stray away from conflating language when describing different security properties. Let's open a separate issue to discuss this further if you want to just so that we don't slow down this PR from being merged.

PR #274, opened 2021-09-02
Make from required

TelegramSam:

In #115 we discussed making anoncrypt not a thing, using disposable DIDs instead. 
In #141 I added that clarification.

I believe the only reason for `from` being optional was for the use of anonymous messaging, which we are no longer directly including. The JWM spec declares it as optional, but it is perfectly acceptable for the use of JWM within DIDComm to impose additional constraints.

Making `from` a required attribute solves the issues brought up with trust ping, and is something I had imaged before but did not apply the full change of removing anoncrypt.

### DIFF
--- a/docs/spec-files/message_structure.md
+++ b/docs/spec-files/message_structure.md
@@ -103,7 +103,7 @@ Headers in DIDComm are intended to be extensible in much the same way that heade
 
 The `to` header cannot be used for routing, since it is encrypted at every intermediate point in a route. Instead, the `forward` message contains a `next` attribute in its body that specifies the target for the next routing operation.
 
-- **from** - OPTIONAL. Sender identifier. The `from` attribute MUST be a string that is a valid DID or [DID URL](https://w3c.github.io/did-core/#did-url-syntax) (without the [fragment component](https://w3c.github.io/did-core/#fragment)) which identifies the sender of the message. When a message is encrypted, the sender key MUST be authorized for encryption by this DID. Authorization of the encryption key for this DID MUST be verified by message recipient with the proper proof purposes. See the [message authentication](#Message-Authentication) section for additional details.
+- **from** - REQUIRED. Sender identifier. The `from` attribute MUST be a string that is a valid DID or [DID URL](https://w3c.github.io/did-core/#did-url-syntax) (without the [fragment component](https://w3c.github.io/did-core/#fragment)) which identifies the sender of the message. When a message is encrypted, the sender key MUST be authorized for encryption by this DID. Authorization of the encryption key for this DID MUST be verified by message recipient with the proper proof purposes. See the [message authentication](#Message-Authentication) section for additional details.
 
   When the sender wishes to be anonymous, they should use a new DID created for the purpose to avoid correlation with any other behavior or identity. Peer DIDs are lightweight and require no ledger writes, and therefore a good method to use for this purpose.
 


kdenhartog (approved):

dhh1128 (approved):
This is fine and could be merged as-is. However, I wonder if one extra sentence might be useful just after "...with the proper proof purposes": "(When operating in anonymous mode, use a new, ephemeral DID.)"

I'm okay merging without this change but just mention it in case it seems helpful.

awoie (approved):
lgtm

awoie:
As per 6 September WG meeting, we want to merge this but @TelegramSam needs to solve the conflicts first.

awoie:
> Signed-off-by: Sam Curren [telegramsam@gmail.com](mailto:telegramsam@gmail.com)
> 
> In #115 we discussed making anoncrypt not a thing, using disposable DIDs instead.
> In #141 I added that clarification.
> 
> I believe the only reason for `from` being optional was for the use of anonymous messaging, which we are no longer directly including. The JWM spec declares it as optional, but it is perfectly acceptable for the use of JWM within DIDComm to impose additional constraints.
> 
> Making `from` a required attribute solves the issues brought up with trust ping, and is something I had imaged before but did not apply the full change of removing anoncrypt.

I'm ok with this PR, but I'd object to removing anoncreds. I do see value in hiding `skid` from the mediator.

docs/spec-files/message_structure.md
```diff
@@ -102,7 +102,7 @@ Headers in [DIDComm Messaging](.) are intended to be extensible in much the same
...
+-**from** - REQUIRED. Sender identifier. The `from` attribute MUST be a string that is a valid [DID](https://w3c.github.io/did-core/) or [DID URL](https://w3c.github.io/did-core/#did-url-syntax) (without the [fragment component](https://w3c.github.io/did-core/#fragment)) which identifies the sender of the message. When a message is encrypted, the sender key MUST be authorized for encryption by this [DID](https://w3c.github.io/did-core/). Authorization of the encryption key for this [DID](https://w3c.github.io/did-core/) MUST be verified by message recipient with the proper proof purposes. When the sender wishes to remain mode, they should use a new, ephemeral DID. See the [message authentication](#Message-Authentication) section for additional details.
```
awoie, line 105: When I first reviewed this PR, I missed the SHOULD (since it is lower case). We should probably discuss this again. Since I also don't want to get rid of anoncreds.

awoie (changes requested):
When I first reviewed this PR, I missed the SHOULD (since it is lower case). We should probably discuss this again. Since I also don't want to get rid of anoncrypt.

kdenhartog:
Just re-reading this now and not sure what's meant by "sender wishes to remain mode". This seems like a colloquialism that I'm not picking up on. @TelegramSam can you explain what you mean with a bit more detail so I can offer a suggestion on how to make this text more clear?

dhh1128:
>"sender wishes to remain mode"

I wonder if this is an artifact of a cut/paste mistake, where the original text said either "wishes to remain anonymous" or "sender is in anonymous mode"?

docs/spec-files/message_structure.md
```diff
@@ -83,7 +83,7 @@ Headers in [DIDComm Messaging](.) are intended to be extensible in much the same
 
 - **id** - REQUIRED. Message ID. The `id` attribute value MUST be unique to the sender.
 
-- **type** - REQUIRED. Plaintext message type, useful for message handling in application-level protocols. The `type` attribute value MUST be a valid [Message Type URI](protocols.md#message-type-uri), that when resolved gives human readable information about the message. The attribute's value SHOULD predict the content in the `body` of the message.
+- **type** - REQUIRED. Plaintext message type, useful for message handling in application-level protocols. The `type` attribute value MUST be a valid [Message Type URI](protocols.md#message-type-uri), that when resolved fgives human readable information about the message. The attribute's value SHOULD predict the content in the `body` of the message.
```
brianorwhatever, line 86:
```suggestion
- **type** - REQUIRED. Plaintext message type, useful for message handling in application-level protocols. The `type` attribute value MUST be a valid [Message Type URI](protocols.md#message-type-uri), that when resolved gives human readable information about the message. The attribute's value SHOULD predict the content in the `body` of the message.
```

PR #277, opened 2021-09-02
Clarifies purpose of `to` attribute

### LINKED ISSUE #210: Questions and comments re routing/forward
Imagine you have the following service in the services section of Bob's DID Doc and this DID Doc was exchanged OOB with Alice:

```
{
    "id": "did:example:bob#didcomm-1",
    "type": "DIDCommMessaging",
    "serviceEndpoint": "did:example:somemediator",
    "routingKeys": ["did:example:anothermediator#somekey"]
}
```
Now, Alice wants to send Bob a message.

Alice would do the following:
- Create the message and pack the message to Bob's `keyAgreement` keys from Bob's DID Doc.
- Dereference `did:example:anothermediator#somekey` and find one particular `keyAgreement` key.
- Create a forward `message A` that wraps the packed message to anothermediator's somekey (JWE with one recipient).
- Resolve `did:example:somemediator` and get somemediator's `keyAgreement` keys as well as somemediator's serviceEndpoint.
- Create another forward `message B` that wraps `message A` to all `keyAgreement` keys of somemediator (JWE with multiple recipients).

Now...

`next` is currently defined as a DID

I assume `message B` looks something like:

```
{
    "type": "https://didcomm.org/routing/2.0/forward",
    "to": ["did:example:somemediator"],
    "expires_time": 1516385931,
    "body":{
        "next": "did:example:anothermediator",
        "payloads~attach": [
            // JWE forward message A
        ]
    }
}
```

And `message A` looks something like:
```
{
    "type": "https://didcomm.org/routing/2.0/forward",
    "to": ["did:example:anothermediator"],
    "expires_time": 1516385931,
    "body":{
        "next": "did:example:bob",
        "payloads~attach": [
            // JWE message from Alice to Bob
        ]
    }
}
```

A couple of questions:
1. The current spec says, don't use `to` for `forward` messages. Is this still the case since the `forward` message examples show the `to` property? A recipient (also mediator) will be able check the `kid` whether it is the intended recipient.
2. Which service endpoint will be used for `message A`? Is Alice expected to get the service endpoint from the DID resolution result of `did:example:anothermediator`? If this is the case, we need normative language for that in the spec.
3. When would you actually have multiple array entries in the `to` property? If Bob has multiple `keyAgreement` keys, would you add the DID URLs of each of those to the `to` property of the message that is intended for Bob? I would not assume so, right (since I expect the recipient section would do that job) ?
4. We should probably add some language that authcrypt for forward messages is discouraged since it would defeat its purpose?

**Updated** to fix some issues in the descripton!

TelegramSam:
Fixes #210

### DIFF
--- a/docs/spec-files/message_structure.md
+++ b/docs/spec-files/message_structure.md
@@ -97,7 +97,7 @@ Headers in DIDComm are intended to be extensible in much the same way that heade
     | Signed | `application/didcomm-signed+json` |
     | Plaintext | `application/didcomm-plain+json`|
 
-- **to** - OPTIONAL. Identifier(s) for recipients. MUST be an array of strings where each element is a valid DID or [DID URL](https://w3c.github.io/did-core/#did-url-syntax) (without the [fragment component](https://w3c.github.io/did-core/#fragment)) that identifies a member of the message's intended audience.
+- **to** - OPTIONAL. Identifier(s) for recipients. MUST be an array of strings where each element is a valid DID or [DID URL](https://w3c.github.io/did-core/#did-url-syntax) (without the [fragment component](https://w3c.github.io/did-core/#fragment)) that identifies a member of the message's intended audience. These values are useful for recipients to know which of their keys can be used for decryption. It is not possible for one recipient to verify that the message was sent to a different recipient.
 
  When Alice sends the same message to Bob and Carol, it is by inspecting this header that Bob and Carol learn that the message was sent to both of them. If the header is omitted, each recipient can only assume they are the only recipient (much like an email sent only to `BCC:` addresses).
 


awoie (approved):
looks good, minor editorial + requested by the WG after multi-recipients discussion

awoie:
@TelegramSam could you please address the conflict, then we can merge (as per 6 September WG meeting)

docs/spec-files/message_structure.md
```diff
@@ -88,6 +88,17 @@ Headers in [DIDComm Messaging](.) are intended to be extensible in much the same
 - **typ** - OPTIONAL. Media type of the JWM content.
 
 - **to** - OPTIONAL. Identifier(s) for recipients. MUST be an array of strings where each element is a valid [DID](https://www.w3.org/TR/did-core/) or [DID URL](https://w3c.github.io/did-core/#did-url-syntax) (without the [fragment component](https://w3c.github.io/did-core/#fragment)) that identifies a member of the message's intended audience.
```
AnomalRoil, line 90: Notice that this PR actually duplicated the **to** field entry, letting the old one above the new part about the table for the media types.

PR #280, opened 2021-09-06
fix: fix build (removed processing_model.md from specs.json)

awoie:
fix: fix build; removed processing_model.md


awoie:
chore: merging this

PR #281, opened 2021-09-06
chore: removed redundant typ table

### LINKED ISSUE #264: Editorial cleanup needed for reconciling typ table in section 3.1 with section 2 sub sections
section 3.1 `typ` describes the various message types that should be included. The section above also describes very similar aspects. This ticket is to track the removal of the table and optionally moving to section 2 and referencing from section 3.1 or completely removing it and making sure the information about the different methods of wrapping are preserved somewhere within section 2.

awoie:
Removing the `typ` table for message types since the information is redundant (also found in specific sections) and the table location makes no sense. Currently, the table is in the plaintext message section but it actually would make more sense to move this to the implementer's guide.

Potentially closes #264

### DIFF
--- a/docs/spec-files/message_structure.md
+++ b/docs/spec-files/message_structure.md
@@ -87,15 +87,6 @@ Headers in [DIDComm Messaging](.) are intended to be extensible in much the same
 
 - **typ** - OPTIONAL. Media type of the JWM content.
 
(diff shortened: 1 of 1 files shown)


kdenhartog (approved):

kdenhartog:
I believe this closes #264 as well

PR #288, opened 2021-09-20
Update message_structure.md

### LINKED ISSUE #266: Plaintext Message "body" content
[Plaintext Message Headers](https://identity.foundation/didcomm-messaging/spec/#message-headers) sections says that
> A DIDComm plaintext message conveys most of its application-level data inside a **JSON** body **object** that is a direct child of the message root. The structure inside body is predicted by the value of the message’s type attribute, and varies according to the definition of the protocol-specific message in question. Each type of message will have its own body.

It's not clear from that description and the provided examples what can be a "body" value:
- JSON Object only (basically key-value pairs), as it says `JSON Object` (see bold above).
- JSON Object or JSON Array
- any valid JSON (JSON object, or JSON array, or JSON value such as a number, string, bool)

In other words, what of the examples below are correct (or all examples are valid):

**Example 1**
```
{
  "type": "https://didcomm.org/my-protocol/1.0",
  "body": {
    "aaa": "bbb"
  }
  ...
}
```

**Example 2**
```
{
  "type": "https://didcomm.org/my-protocol/1.0",
  "body": "aaa"
  ...
}
```

**Example 3**
```
{
  "type": "https://didcomm.org/my-protocol/1.0",
  "body": 42
  ...
}
```

**Example 4**
```
{
  "type": "https://didcomm.org/my-protocol/1.0",
  "body": [
   {"aaa": "bbb"}, {"ccc": "ddd"}
  ]
  ...
}
```

brianorwhatever:
Clarifying what the body attribute looks like
closes #266

### DIFF
--- a/docs/spec-files/message_structure.md
+++ b/docs/spec-files/message_structure.md
@@ -116,7 +116,7 @@ The `to` header cannot be used for routing, since it is encrypted at every inter
 
 - **expires_time** - OPTIONAL. Message Expired Time. The `expires_time` attribute is used for the sender to express when they consider the message to be expired, expressed in UTC Epoch Seconds (seconds since 1970-01-01T00:00:00Z UTC) [link](1970-01-01T00:00:00Z UTC). This attribute signals when the message is considered no longer valid by the sender. When omitted, the message is considered to have no expiration by the sender.
 
-- **body** - REQUIRED. The `body` attribute contains all the message type specific attributes of the message type indicated in the `type` attribute. This attribute MUST be present, even if empty.
+- **body** - REQUIRED. The `body` attribute contains all the message type specific attributes of the message type indicated in the `type` attribute. This attribute MUST be present, even if empty. It MUST be a JSON object conforming to RFC 7159.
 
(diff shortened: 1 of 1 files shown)


docs/spec-files/message_structure.md
```diff
@@ -116,7 +116,7 @@ The `to` header cannot be used for routing, since it is encrypted at every inter
 
 - **expires_time** - OPTIONAL. Message Expired Time. The `expires_time` attribute is used for the sender to express when they consider the message to be expired, expressed in UTC Epoch Seconds (seconds since 1970-01-01T00:00:00Z UTC) [link](1970-01-01T00:00:00Z UTC). This attribute signals when the message is considered no longer valid by the sender. When omitted, the message is considered to have no expiration by the sender.
 
-- **body** - REQUIRED. The `body` attribute contains all the message type specific attributes of the message type indicated in the `type` attribute. This attribute MUST be present, even if empty.
+- **body** - REQUIRED. The `body` attribute contains all the message type specific attributes of the message type indicated in the `type` attribute. This attribute MUST be present, even if empty. It MUST be a JSON object conforming to RFC 7159.
```
kdenhartog, line 119:
```suggestion
- **body** - REQUIRED. The `body` attribute contains all the message type specific attributes of the message type indicated in the `type` attribute. This attribute MUST be present, even if empty. It MUST be a JSON object conforming to [RFC 7159](https://datatracker.ietf.org/doc/html/rfc7159).
```

Issue #289, opened 2021-09-21
Signed Messages needs to require the `to` header


AnomalRoil:
Currently the "to" field is optional, but when doing "sign then encrypt", it is a good practice to **mandate** the use of a signed "to" field, in order to avoid tricky issues and attacks.

See e.g. [world.std.com/~dtd/sign_encrypt/sign_encrypt7.html](https://theworld.com/~dtd/sign_encrypt/sign_encrypt7.html)  for more details about what could go wrong.

Currently the specs says:

> When a message is both signed and encrypted, the plaintext is signed, and then the signed envelope is encrypted. The opposite order is not used, since it would imply that the signer committed to opaque data (which is unsafe and undermines non-repudiation).

(Notice this isn't using requirement terminology, we might want to use a "The opposite order SHALL NOT be used" instead?)

And also:

> A DIDComm message can be signed, either in conjunction with encryption or independently if the message will remain unencrypted.
> 
> If a message is signed and encrypted to add non-repudiation, it must be signed prior to encryption.

Let's make the "to" field mandatory when we are doing "sign then encrypt" in order to avoid relay attacks or replay attacks or other nasty issues.

A few more things:
- For a signed broadcast message, we might want to mandate a "created_time" timestamp instead and a "to" field that says it is a broadcast? Or should we add in the "verification" section that a message that is lacking a "to" field and is signed should be automatically treated as a broadcast?
- Arguably, authcrypt isn't vulnerable to "surreptitious forwarding" since it's authenticating the sender, but in the anonymous case, or with an ephemeral DID, this isn't the case... So I'd argue it's still a concern.
- Encrypt then sign requires the "from" attribute to be encrypted, which is now required as of #274, so we're fine on that front, although it's currently not supported by the spec, and so it's currently more secure than the Sign then Encrypt we're mandating...

kdenhartog:
In a sign then encrypt (nested) message I believe we're covered here because the from value should match to authorized keys in the did document that were used for signing and key agreement.

If we were to only sign the message then we'd encounter this issue as well, but there's certain cases where it makes sense to sign with no intended audience such as signed proof requests in a 0-RTT flow for VC-Authn. 

For this reason, I think it makes sense to stick with a SHOULD in this scenario where implementations should be including the to value, but there's good reasons not to in certain cases and we can detail this should in the implementation guide.

AnomalRoil:
[...]

PR #296, opened 2021-09-28
Recommend the usage of the `to` header for signed messages

AnomalRoil:
As discussed in #289, we need to have a properly defined `to` header when signing messages, otherwise forwarded messages could wrongly be trusted as being intended for us when they are not.
This also means we must verify the inclusion of ourself in the `to` header upon reception of a signed message, and warn otherwise.

Also this is removing the duplicated entry for `to` that was wrongly introduced in #277.

### DIFF
--- a/docs/spec-files/message_structure.md
+++ b/docs/spec-files/message_structure.md
@@ -32,7 +32,7 @@ The [media type](https://tools.ietf.org/html/rfc6838) of a **DIDComm signed mess
 
 The media type of the envelope MUST be set in the `typ` [property](https://tools.ietf.org/html/rfc7515#section-4.1.9) of the JWS.
 
-In order to avoid [surreptitious forwarding or malicious usage](https://theworld.com/~dtd/sign_encrypt/sign_encrypt7.html) of a signed message, a **DIDComm signed message** SHOULD contain a properly defined `to` header, In the case when a message is *both* signed and encrypted the inner JWM being signed MUST contain a `to` header.
+In order to avoid [surreptitious forwarding or malicious usage](https://theworld.com/~dtd/sign_encrypt/sign_encrypt7.html) of a signed message, a **DIDComm signed message** SHOULD contain a properly defined `to` header. In the case where a message is *both* signed and encrypted, the inner JWM being signed MUST contain a `to` header.
 
 When persisted as a file or attached as a payload in other contexts, the file extension for **DIDComm signed messages** SHOULD be `dcsm`, giving a globbing pattern of `*.dcsm`; this SHOULD be be read as "Star Dot D C S M" or as "D C S M" files. A possible icon for this media type depicts a signed envelope ([svg](../collateral/dcsm.svg) | [256x256](../collateral/dcsm-256.png) | [128x128](../collateral/dcsm-128.png) | [64x64](../collateral/dcsm-64.png)):
 
@@ -88,6 +88,8 @@ Headers in [DIDComm Messaging](.) are intended to be extensible in much the same
 - **type** - REQUIRED. Plaintext message type, useful for message handling in application-level protocols. The `type` attribute value MUST be a valid [Message Type URI](protocols.md#message-type-uri), that when resolved gives human readable information about the message. The attribute's value SHOULD predict the content in the `body` of the message.
 
 - **typ** - OPTIONAL. Media type of the JWM content.
+
+- **to** - OPTIONAL. Identifier(s) for recipients. MUST be an array of strings where each element is a valid [DID](https://www.w3.org/TR/did-core/) or [DID URL](https://w3c.github.io/did-core/#did-url-syntax) (without the [fragment component](https://w3c.github.io/did-core/#fragment)) that identifies a member of the message's intended audience.
     
     The following table provides an overview of using the media type properties in supported [DIDComm Messaging](.) messages:
 
@@ -102,9 +104,9 @@ Headers in [DIDComm Messaging](.) are intended to be extensible in much the same
 
  When Alice sends the same message to Bob and Carol, it is by inspecting this header that Bob and Carol learn that the message was sent to both of them. If the header is omitted, each recipient can only assume they are the only recipient (much like an email sent only to `BCC:` addresses).
 
-For signed messages, there's specific requirements around properly defined `to` header outlined in the `to` header definition above. The reason for this is to prevents certain kind of [forwarding attacks](https://theworld.com/~dtd/sign_encrypt/sign_encrypt7.html), where a message that wasn't meant for a given recipient is forwarded along with its signature to that recipient which then could blindly trust it because of the signature.
+For signed messages, there are specific requirements around properly defining the `to` header outlined in the **DIDComm Signed Message** definition above. The reason for this is to prevents certain kind of [forwarding attacks](https://theworld.com/~dtd/sign_encrypt/sign_encrypt7.html), where a message that was not meant for a given recipient is forwarded along with its signature to a recipient which then could blindly trust it because of the signature.
 
-Upon reception of a message whose `to` header is defined, the recipient SHOULD verify that they are included in that field and implementations SHOULD give a warning to their user when it is not the case as it could indicate malicious intent from the sender.
+Upon reception of a message whose `to` header is defined, the recipient SHOULD verify that they are included in that field. Implementations MUST NOT fail when it is not the case and SHOULD give a warning to their user as it could indicate malicious intent from the sender.
 
 The `to` header cannot be used for routing, since it is encrypted at every intermediate point in a route. Instead, the `forward` message contains a `next` attribute in its body that specifies the target for the next routing operation.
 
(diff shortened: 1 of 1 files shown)


docs/spec-files/message_structure.md
```diff
@@ -26,12 +26,14 @@ A **DIDComm signed message** is a signed [JWM (JSON Web Messages)](https://tools
...
 
-When a message is *both* signed and encrypted, the plaintext is signed, and then the signed envelope is encrypted. The opposite order is not used, since it would imply that the signer committed to opaque data (which is unsafe and undermines non-repudiation).
+When a message is *both* signed and encrypted, the plaintext SHOULD be signed, and then the signed envelope SHOULD be encrypted. The opposite order SHOULD NOT be used, since it would imply that the signer committed to opaque data (which is unsafe and undermines non-repudiation).
```
kdenhartog, line 29:
```suggestion
When a message is *both* signed and encrypted, the plaintext MUST be signed, and then the signed envelope MUST be encrypted. The opposite order MUST NOT be used, since it would imply that the signer committed to opaque data (which is unsafe and undermines non-repudiation).
```

I see no justifiable reason to do the opposite ordering, so I'd prefer to cut that code path out with a MUST here.

docs/spec-files/message_structure.md
```diff
@@ -26,12 +26,14 @@ A **DIDComm signed message** is a signed [JWM (JSON Web Messages)](https://tools
...
 
 The [media type](https://tools.ietf.org/html/rfc6838) of a **DIDComm signed message** MUST be `application/didcomm-signed+json`.
 
 The media type of the envelope MUST be set in the `typ` [property](https://tools.ietf.org/html/rfc7515#section-4.1.9) of the JWS.
 
+In order to avoid [surreptitious forwarding or malicious usage](https://theworld.com/~dtd/sign_encrypt/sign_encrypt7.html) of a signed message, a **DIDComm signed message** SHOULD contain a properly defined `to` header, this is also the case when a message is *both* signed and encrypted.
```
kdenhartog, line 35:
```suggestion
In order to avoid [surreptitious forwarding or malicious usage](https://theworld.com/~dtd/sign_encrypt/sign_encrypt7.html) of a signed message, a **DIDComm signed message** SHOULD contain a properly defined `to` header, In the case when a message is *both* signed and encrypted the inner JWM being signed MUST contain a `to` header.
```

docs/spec-files/message_structure.md
```diff
@@ -86,8 +88,6 @@ Headers in [DIDComm Messaging](.) are intended to be extensible in much the same
 - **type** - REQUIRED. Plaintext message type, useful for message handling in application-level protocols. The `type` attribute value MUST be a valid [Message Type URI](protocols.md#message-type-uri), that when resolved gives human readable information about the message. The attribute's value SHOULD predict the content in the `body` of the message.
 
 - **typ** - OPTIONAL. Media type of the JWM content.
-
-- **to** - OPTIONAL. Identifier(s) for recipients. MUST be an array of strings where each element is a valid [DID](https://www.w3.org/TR/did-core/) or [DID URL](https://w3c.github.io/did-core/#did-url-syntax) (without the [fragment component](https://w3c.github.io/did-core/#fragment)) that identifies a member of the message's intended audience.
```
kdenhartog, line 90: I'd prefer to see us move the test into this section here to remain consistent in our definition of headers rather than move the normative requirements into a separate section which doesn't call out that it's making requirements around the usage of the `to` header.

docs/spec-files/message_structure.md
```diff
@@ -102,6 +102,10 @@ Headers in [DIDComm Messaging](.) are intended to be extensible in much the same
 
  When Alice sends the same message to Bob and Carol, it is by inspecting this header that Bob and Carol learn that the message was sent to both of them. If the header is omitted, each recipient can only assume they are the only recipient (much like an email sent only to `BCC:` addresses).
 
+For signed messages, notice that it is RECOMMENDED to have a properly defined `to` header, since it prevents certain kind of [forwarding attacks](https://theworld.com/~dtd/sign_encrypt/sign_encrypt7.html), where a message that wasn't meant for a given recipient is forwarded along with its signature to that recipient which then could blindly trust it because of the signature.
```
kdenhartog, line 105:
```suggestion
For signed messages, there's specific requirements around properly defined `to` header outlined in the `to` header definition above. The reason for this is to prevents certain kind of [forwarding attacks](https://theworld.com/~dtd/sign_encrypt/sign_encrypt7.html), where a message that wasn't meant for a given recipient is forwarded along with its signature to that recipient which then could blindly trust it because of the signature.
```

kdenhartog: Changing this so that someone reading this section knows to reference the other section for requirements so we're not redundantly stating normative statements.

docs/spec-files/message_structure.md
```diff
@@ -102,6 +102,10 @@ Headers in [DIDComm Messaging](.) are intended to be extensible in much the same
...
+
+Upon reception of a message whose `to` header is defined, the recipient SHOULD verify that they are included in that field and implementations SHOULD give a warning to their user when it is not the case as it could indicate malicious intent from the sender.
```
kdenhartog, line 107: I'm good with these SHOULDs here

docs/spec-files/message_structure.md
AnomalRoil: This is just removing the duplicated entry for the "to" header. This was duplicated in #277 without reason.

I think it was a copy pasta mistake.

If you look below in that file, it's still containing this exact same paragraph.

docs/spec-files/message_structure.md
AnomalRoil: I could see protocols where you have a signature on a ciphertext, in order to allow anyone to verify the provenance of the ciphertext without allowing decryption...

Typically you could do Sign then Encrypt then Mac/Sign... Having a strong requirement here seemed a bit restrictive. 

Also anything that is based on homomorphic encryption might want to not embed the signature within its ciphertext... Arguably, the structure of the plaintext wouldn't allow for such schemes with DIDComm, so it's probably not a problem to rule out such possibilities.

docs/spec-files/message_structure.md
AnomalRoil: It seems @TelegramSam is also fixing this in his latest PR #274.
So I guess I can drop this removal from this PR.

docs/spec-files/message_structure.md
```diff
@@ -26,12 +26,14 @@ A **DIDComm signed message** is a signed [JWM (JSON Web Messages)](https://tools
...
 
-When a message is *both* signed and encrypted, the plaintext is signed, and then the signed envelope is encrypted. The opposite order is not used, since it would imply that the signer committed to opaque data (which is unsafe and undermines non-repudiation).
+When a message is *both* signed and encrypted, the plaintext SHOULD be signed, and then the signed envelope SHOULD be encrypted. The opposite order SHOULD NOT be used, since it would imply that the signer committed to opaque data (which is unsafe and undermines non-repudiation).
 
 The [media type](https://tools.ietf.org/html/rfc6838) of a **DIDComm signed message** MUST be `application/didcomm-signed+json`.
 
 The media type of the envelope MUST be set in the `typ` [property](https://tools.ietf.org/html/rfc7515#section-4.1.9) of the JWS.
 
+In order to avoid [surreptitious forwarding or malicious usage](https://theworld.com/~dtd/sign_encrypt/sign_encrypt7.html) of a signed message, a **DIDComm signed message** SHOULD contain a properly defined `to` header. In the case where a message is *both* signed and encrypted, the inner JWM being signed MUST contain a `to` header.
```
AnomalRoil, line 35: @kdenhartog Notice that here we are recommending (SHOULD) usage of the header for "just" signed messages, but mandating (MUST) it for signed and encrypted message, as per your suggested change. That's fine for me, but I just wanted to double check with you that's what you intended.

docs/spec-files/message_structure.md
kdenhartog: I don't see value in adding that additional code path here. I'd prefer to leave these problems out of scope here since it goes against the [recommendations of JOSE](https://datatracker.ietf.org/doc/html/rfc7519#section-11.2). I'd prefer to remain in alignment with this rather than support these additional aspects being suggested here.

docs/spec-files/message_structure.md
kdenhartog: Thanks for pointing that out, I didn't realize this.

docs/spec-files/message_structure.md
kdenhartog: Yeah that's what I was thinking for this. This looks good.

docs/spec-files/message_structure.md
AnomalRoil: How about not using normative terminology and linking to that JOSE section?

AnomalRoil: @kdenhartog ?

AnomalRoil:
I believe I've changed everything as we discussed. Let me know what you think now.

AnomalRoil:
@kdenhartog ^

TelegramSam:
Merged after discussion in WG 20220110.

Issue #292, opened 2021-09-21
Why is P-256 deprecated?


AnomalRoil:
I've already touched on the topic in #243, especially in https://github.com/decentralized-identity/didcomm-messaging/issues/243#issuecomment-890029722 but the rest of the issue hijacked the question.

Mandating P-384 and deprecating P-256 over P-384 doesn't necessarily make sense in my opinion.
X25519 and P-256 both have a security level of ~120-128 bits.
P-384 has a security level of ~192 bits.
So it doesn't make sense to deprecate P-256 but not X25519... If you want a larger security level you should also deprecate X25519.
P-384 is also far less used and thus optimized in software and hardware implementations, which means it is slower than the more classical P-256 curve.

I've already posted many benchmarks in #243, but let me re-add this one:
```
p256::generate_key_pair                  bench:   9,665 ns/iter
p256::gen_key_pair_and_agree_ephemeral   bench:  51,820 ns/iter

p384::generate_key_pair                  bench: 362,917 ns/iter
p384::gen_key_pair_and_agree_ephemeral   bench: 708,025 ns/iter

x25519::generate_key_pair                bench:  19,816 ns/iter
x25519::gen_key_pair_and_agree_ephemeral bench:  73,051 ns/iter
```

I don't see the rationale behind this choice and wouldn't recommend it. Actually removing P-384 would make things simpler, since it's not an algorithm that is necessarily available in all libraries of platforms, unlike P-256.
(But it would limit DIDComm to a 128 bit level, agreed, which is not necessarily desirable neither.)

The security level of all the symmetric encryption primitives is at 256 bits, so we're only limited by the key exchange currently. But key-exchange is a CPU intensive operation, especially at higher security levels.

I would also recommend adding a section about allowing the use of a pre-shared key, so that DIDComm is able to reach its complete 256 bit security level using out-of-band key sharing, unlike with any current key-agreement option. This is already discussed in #255.  
This would also ease the implementation of "session" on top of DIDComm. (As per § 1.3.2, "sessions can _easily_ be built atop it", which isn't the case currently as it would require designing an alternative protocol to store the ciphertext in the body part of a DIDComm plaintext message, or something similarly strange...)

So, why P-384 as a default and not P-256?
I would recommend against deprecating P-256.

kdenhartog:
[...]

PR #299, opened 2021-10-04
DIDComm Test Vectors

asmarty:


docs/spec-files/appendix.md
```diff
@@ -0,0 +1,671 @@
+## Appendix
+This section provides test vectors of DIDComm Messages. The test vectors defined in the [Appendix C](#appendix-c-test-vectors) MIGHT be used to validate implementations of [DIDComm Messaging](.). The test vectors  MUST use the sender and recipient private keys defined in the [Appendix A](#appendix-a-secrets-for-test-vectors) to encrypt and sign DIDComm Messages. The test vectors MUST use the sender and recipient public keys defined in the [Appendix B](#appendix-b-diddocs-for-test-vectors) to decrypt DIDComm Encrypted Messages and verify signature of DIDComm Signed Messages.
```
AnomalRoil, line 2: > The test vectors  MUST use the sender and recipient **private keys** defined in the [Appendix A](#appendix-a-secrets-for-test-vectors) to **encrypt** and sign DIDComm Messages 

NB: a private key is used to decrypt, not to encrypt.

> The test vectors MUST use the sender and recipient **public keys** defined in the [Appendix B](#appendix-b-diddocs-for-test-vectors) to **decrypt** DIDComm Encrypted Messages and verify signature of DIDComm Signed Messages.

And a public key is used to encrypt, not to decrypt, when talking about encryption.

docs/spec-files/appendix.md
```diff
@@ -0,0 +1,671 @@
...
+            "kid":"did:example:bob#key-x25519-3"
+         }
+      }
+   ],
+   "tag":"uYeo7IsZjN7AnvBjUZE5lNryNENbf6_zew_VC-d4b3U",
+   "iv":"o02OXDQ6_-sKz2PX_6oyJg"
+}
+```
+
+In this example, the message is first signed with EdDSA digital signature and then encrypted with ECDH-1PU key wrapping algorithm using key with NIST defined P-521 elliptic curve and A256CBC-HS512 for content encryption of the message.
```
AnomalRoil, line 623: I believe as per #292 discussion and other discussion that P-521 isn't meant to be supported. We might want to only include examples using P-256 and P-384 and X25519 instead.

docs/spec-files/appendix.md
```diff
@@ -0,0 +1,671 @@
...
+            "kid":"did:example:bob#key-p384-2"
+         }
+      }
+   ],
+   "tag":"bkodXkuuwRbqksnQNsCM2YLy9f0v0xNgnhSUAoFGtmE",
+   "iv":"aE1XaH767m7LY0JTN7RsAA"
+}
+```
+
+This example uses ECDH-ES key wrapping algorithm using key with NIST defined P-521 elliptic curve and A256GCM for content encryption of the message.
```
AnomalRoil, line 567: I believe as per #292 discussion and other discussion that P-521 isn't meant to be supported. We might want to only include examples using P-256 and P-384 and X25519 instead.

docs/spec-files/appendix.md
```diff
@@ -0,0 +1,671 @@
...
+            "d":"pFRUKkyzx4kHdJtFSnlPA9WzqkDT1HWV0xZ5OYZd2SY",
+            "crv":"Ed25519",
+            "x":"G-boxFB6vOZBu-wXkm-9Lh79I8nf9Z50cILaOgKKGww"
+         }
+      }
+   },
+   {
+      "kid":"did:example:alice#key-2",
+      "type":"JsonWebKey2020",
+      "verificationMaterial":{
```
OR13, line 30: Im not sure what the type of this is... but i sorta wish it , looked like the keys here:

```
  {
    "id": "did:key:z6MkptB2HAniuXsVANcR938iRdkq18GXE4c7uaJPjdp9QFY7#z6MkptB2HAniuXsVANcR938iRdkq18GXE4c7uaJPjdp9QFY7",
    "type": "JsonWebKey2020",
    "controller": "did:key:z6MkptB2HAniuXsVANcR938iRdkq18GXE4c7uaJPjdp9QFY7",
    "publicKeyJwk": {
      "kty": "OKP",
      "crv": "Ed25519",
      "x": "mva13N9iYbjwomoGaEtMzNvmOpoNPj0uuSojznBDh3w"
    },
    "privateKeyJwk": {
      "kty": "OKP",
      "crv": "Ed25519",
      "x": "mva13N9iYbjwomoGaEtMzNvmOpoNPj0uuSojznBDh3w",
      "d": "uxxS11jjJ5JG4V2uH1cz-RqMpJoAy66c2w3M_vuCZl0"
    }
  }
 ```
 
 https://did.key.transmute.industries/generate/ed25519?seed=bb1c52d758e3279246e15dae1f5733f91a8ca49a00cbae9cdb0dccfefb82665d
 
 If it's possible, we should avoid making new data structures.

File: docs/spec-files/appendix.md (hunk 3 above)
Comment (OR13) [reply docs/spec-files/appendix.md:567]: yeah i would avoid P-521, its not well supported.

**Review by OR13 — APPROVED** [review]
Overall these look good, my only request would be to avoid representing private keys so differently from public keys, but I am maybe missing the history of that decision.

**Comment by OR13** [comment]
Would this be legal?

```
{
      "kid":"did:example:alice#key-1",
      "type":"JsonWebKey2020",
      "verificationMaterial":{
         "format":"Base58",
         "value": " ..."
   },
```

^ this would seem to to be encouraged by inventing a new representation for private keys.

File: docs/spec-files/appendix.md [hunk 5]
```diff
@@ -0,0 +1,671 @@
+## Appendix
```
kdenhartog, line 1: Appendix's traditionally are used to add additional information and are therefore considered non-normative. Additionally, the purpose of normative statements have most often been used to call out what's testable so making normative statements about how to test something is a bit recursive.

We should remove any normative statements in this portion.

kdenhartog:
I do generally like the direction this is going. Given that DIDComm is expected to be transport agnostic I think the test vector approach makes more sense for testing compliance and may help to reduce the tension that was encountered when the test harness was built for the Aries project.

PR #302, opened 2021-10-05
Update Version of Trust Ping

TelegramSam:
The version of the trust ping protocol was listed as 1.0, but it's different than the 1.0 defined in the Aries community. This PR changes the vesion to avoid that confusion.


awoie:
As per 11 Oct WG meeting, merging this.

PR #313, opened 2021-11-15
Update Discover Features to 2.0

TelegramSam:


swcurran:
Shouldn't this be V3.0?  V2.0 is in Aries RFCs [RFC 0557](https://github.com/hyperledger/aries-rfcs/tree/main/features/0557-discover-features-v2).

TelegramSam:
this is identical to the aries defined one. other protocols needed version changes because of attachments etc,, but this one has none (yet).

swcurran:
That's fun -- two "identical" specs.  Hmmm...

kdenhartog:
They don't look quite identity to me unless the `queries` and `disclosures` properties are going to be considered headers that we declare here. Technically because they've been moved into the body, it's a breaking change on the protocol even though it's a very small one.

swcurran:
Good point.  Breaking change because we're deleting two fields and adding two.  Having them go in the header does not make any sense -- that's just breaking DIDComm v2.

I think that confirms that every protocol will need a major version update -- decorators or not.

docs/spec-files/feature_discovery.md
```diff
@@ -52,15 +52,15 @@ A `discover-features/query` message looks like this:
...
 
-Implementations of this protocol must recognize the following values for `feature-type`: `protocol`, `goal-code`, `gov-fw`, `didcomm-version`, and `decorator`/`header`. (The concept known as `decorator` in DIDComm v1 approximately maps to the concept known as `header` in DIDComm v2. The two values should be considered synonyms and must both be recognized.) Additional values of `feature-type` may be standardized by raising a PR against this RFC that defines the new type and increments the minor protocol version number; non-standardized values are also valid, but there is no guarantee that their semantics will be recognized.
+Implementations of this protocol must recognize the following values for `feature-type`: `protocol`, `goal-code`, `gov-fw`, `didcomm-version`, and `header`.  Additional values of `feature-type` may be used, and unrecognized values MUST be ignored.
```
kdenhartog, line 63: I think we should state the purpose of each of these values if we're going to state they must be recognized.

docs/spec-files/feature_discovery.md
```diff
@@ -85,7 +85,7 @@ A `discover-features/disclose` message looks like this:
             },
             {
                 "feature-type": "goal-code",
-                "id": "aries.sell.goods.consumer"
+                "id": "org.didcomm.sell.goods.consumer"
```
kdenhartog, line 88: These probably need to be converted into URI based identifiers if they're going to need to be standardized on and useful for the processing of messages without causing name collision issues. What you're suggesting could be the syntax of the URI we use if we want but we'll likely want to define the structure of this.

docs/spec-files/general_messaging_constructs.md
```diff
@@ -10,6 +10,14 @@
 * Errors
 * Tracing / troubleshooting
 
+## Goal Codes
+
+Goal codes are used to coordinate the purpose of an interaction. Some protocols are generic enough to be used for different purposes, and goal codes are a mechanism to communicate the purpose of an interaction in that situation. They are used in Out of Band messages for that reason.
```
kdenhartog, line 15: When I re-read through this I keep getting caught on the purpose of these. I think it's because we're coupling the semantics of human meaningful interactions to a particular identifier so that we can infer how to process the interaction at a technical level and that seems off to me when we've already essentially got that in the protocol identifiers. Essentially I remain confused on when I should be differentiating processing based on a goal code rather than based on the protocol in use.

From this description it sounds like goal codes are being used as an encapsulation mechanism to describe a [coprotocol](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0478-coprotocols) and I'm wondering why a coprotocol URI couldn't be defined and to drop this.

swcurran:
Co-protocols and goal codes are different beasts AFAIK.  They are to use multiple protocols to accomplish a single outcome.  A goal-code is to convey what goal we're trying to accomplish. That is stated in the co-protocols RFC.  At this point coprotocols is a concept not a feature and the RFC just says that implementations should expect that an outcome may require some sort of orchestration of a number of protocols.  There is not a way to express (AFAIK) -- I want to do orchestration X with you, which requires executing Y and Z in some sequence. Further, the RFC implies that if that does become a feature "X" will be a goal-code.

docs/spec-files/feature_discovery.md
TelegramSam: discussed - language needs to be softened around goal code, maybe rmove gov-fw.

docs/spec-files/feature_discovery.md
TelegramSam: I'm pretty resistant to going to that level. Experience in DIDComm v1 does not signal it to be a problem, and moving to non-conflicting values in the future is a non-breaking change.

kdenhartog: That would depend on the implementation. If I decide to only check that the value is a string and then the spec moves to URIs then you're correct it's not a breaking change. However, if I decide to use a regex to validate it's a reverse package notation and then a URI is used it would be an issue.

I'm not gonna die on this hill if you're adamant in not changing it, but I do think it's a mistake that will subtlety sneak up on us at scale and force us to build a centralized namespace registry to avoid conflicts here.

swcurran: @kdenhartog -- you are hereby granted full access to an "I-told-you-so" token to place in the collective face of the WG. You may play your token at any time.

:-)

kdenhartog: I've been waiting to earn one of those elusive tokens 😄

PR #316, opened 2021-12-13
make pthid consistant in oob

### LINKED ISSUE #304: OOB id becomes thid or pthid? Inconsistency.
Comment made by @swcurran on #300:
> While not in the changes made here, there is inconsistency in what the OOB id should become in the response message -- thid or pthid. In Aries RFC 434, the id becomes the pthid, to enable supporting multi-use invitations. This is removed in one part of this proposed protocol, but remains in place in a second part. Which is correct? My vote would be to keep with the pthid, but I've missed any discussions in the DIDComm V2 working group on changing that.
(Comment moved here to allow the unrelated PR to continue.

TelegramSam:
Fixes #304

### DIFF
--- a/docs/spec-files/out_of_band.md
+++ b/docs/spec-files/out_of_band.md
@@ -14,7 +14,7 @@ Second, any identifiers passed in a message sent via URL or QR code must no long
 
 ##### Message Correlation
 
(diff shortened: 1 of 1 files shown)


TelegramSam:
Merged after discussion in WG 20211213.

PR #320, opened 2022-01-07
added transport requirements

### LINKED ISSUE #222: need repeatable methodology for new transports
We have info about how we do DIDComm over HTTP. I want a recipe for how to write up instructions for new transports. Are these instructions called a "note", and "extension", a "transport plugin", or what? Where do I publish one, and what should it contain?

The most substantial effort like this that I know about is [the one for BLE](https://github.com/decentralized-identity/didcomm-bluetooth/blob/main/spec.md). I want to do another for SMTP, another for raw sockets, another for NFC, and another for SMTP.

TelegramSam:
Fixes #222 


TelegramSam:
Kyle to propose clarifications to service endpoint definitions (WG 20220110

mccown:

PR #307, opened 2021-10-12
Add info about Perfect Forward Secrecy

dhh1128:

PR #327, opened 2022-02-01
fix #245

### LINKED ISSUE #245: Need to explicitly define security requirements of DIDCommV2
After reading through #243 entirely, I've come to realize it's probably best if we get @AnomalRoil to help us out with defining explicit security requirements for the **envelope layer only**. I think by doing this we'll be able to identify potential conflicts that we never noticed as well as explicitly define the architectural security requirements we have. So far these are the ones that pop into my head.

MUST have weak perfect forward secrecy. SHOULD have Perfect forward secrecy.
MUST have message confidentiality
MUST have message integrity
MUST have message authentication (In the traditional cryptographic definition used - basically needs to use Authenticated encryption)
MUST default to using repudiable identifier authentication at the envelope layer (authcrypt style "message authentication" - how we've traditionally used the terminology)
MUST be able to add non-repudiable identifier authentication at the envelope layer (digital signatures - this is what nesting gets us)

Note, I'm explicitly saying "identifier" rather than "identity" here. What I mean here is "identifier authentication" is that the recipient can verify continuity of the identifier. When I say "identity authentication" I mean the higher order binding of an identifier to other attributes which is traditional described as "exchange some VCs to prove who you are" which I considered out of scope for this part. It's necessary for the "prevent MITM" part - but we could still perform some formal verification at this layer by making this discrepancy by adding an assumption "Identifier has been bound to a trusted identity" to the verification engine.

I'm sure there are other ones that I'm missing. @dhh1128 @andrewwhitehead @Baha-sk @awoie can you add any ones that I've missed - even if they may be contentious so we can start with an all encompassing list and remove as necessary.

@vimmerru @jovfer @ashcherbakov @devin-fisher @swcurran @TelegramSam @nage @mikelodder7 @SvenHammann90 thinking you guys might have some to provide as well.

dhh1128:

### DIFF
--- a/docs/spec-files/requirements.md
+++ b/docs/spec-files/requirements.md
@@ -2,8 +2,8 @@
 
 The [DIDComm Messaging](.) design attempts to be:
 
-1. **Secure** (specifically, MUST preserve the confidentiality of messages against eavesdropping, MUST preserve the integrity of messages against tampering, and MUST allow the authenticity of messages and message senders to be proved; MUST use best-of-breed crypto; MUST allows parties to emit both repudiable and non-repudiable messages; perfect forward secrecy is not formally required due to the lack of a session construct, but similar outcomes must be achievable, as described in the [Implementer's Guide](guide.md#pfs))
-2. **Private** (unauthorized third parties can’t learn who’s communicating about what, when; lets sender be anonymous to recipient) 
+1. **Secure** (specifically, MUST preserve the integrity of messages against tampering, and MUST allow the authenticity of messages and message senders to be proved; MUST use best-of-breed crypto; MUST allows parties to emit both repudiable and non-repudiable messages; perfect forward secrecy is not formally required due to the lack of a session construct, but similar outcomes must be achievable, as described in the [Implementer's Guide](guide.md#pfs))
+2. **Private** (MUST prevent unauthorized third parties from learning who’s communicating about what, when; MUST let sender be anonymous to recipient) 
 3. **Decentralized** (derives trust for encryption, signing, authn, and authz from control of [DIDs]((https://www.w3.org/TR/did-core/)) rather than oracles like CAs, IDPs, etc; usable at the edge)
 4. **Transport-agnostic** (usable over HTTPS 1.x and 2.0, WebSockets, BlueTooth, chat, push notifications, libp2p, AMQP, SMTP, NFC, sneakernet, snail mail; supports both simplex and duplex; works offline; doesn't assume client-server or synchronous or real-time; allows paired or n-wise or public broadcast usage)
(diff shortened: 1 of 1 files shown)


docs/spec-files/requirements.md
```diff
@@ -2,7 +2,7 @@
 
 The [DIDComm Messaging](.) design attempts to be:
 
-1. **Secure** (tamper-proof; uses best-of-breed crypto; allows parties to talk both on and off the record.)
+1. **Secure** (specifically, MUST have message confidentiality,message integrity, and message authentication; MUST use best-of-breed crypto; MUST allows parties to emit both repudiable and non-repudiable messages.)
```
vimmerru, line 5: I think we need explain `Confidentiality` more here. What exactly it means.

Just an example. We define non-repudiation layer without clear message is it explicitly intended for non-repudiation or in some cases for auth too. Unintended non-repudiation looks like confidentiality issue.

vimmerru, line 5: It would be fair to say about current state of forward secrecy as it often assumed in *secure* protocols.

dhh1128: There is already a separate PR about PFS; see #307. It is in the Implementer's Guide rather than the spec, because PFS is not a requirement. But I agree it is relevant here. I have just added an additional update to the text here to mention it briefly; if/when the PFS PR is merged, we can update this with a better link.

dhh1128: Okay, I have updated the language in an attempt to make it clearer. Do you like it better now?

docs/spec-files/requirements.md
```diff
@@ -2,7 +2,7 @@
 
 The [DIDComm Messaging](.) design attempts to be:
 
-1. **Secure** (tamper-proof; uses best-of-breed crypto; allows parties to talk both on and off the record.)
+1. **Secure** (specifically, MUST preserve the confidentiality of messages against eavesdropping, MUST preserve the integrity of messages against tampering, and MUST allow the authenticity of messages and message senders to be proved; MUST use best-of-breed crypto; MUST allows parties to emit both repudiable and non-repudiable messages; perfect forward secrecy is not formally required due to the lack of a session construct, but similar outcomes must be achievable, as described in the [Implementer's Guide](guide.md#pfs))
```
vimmerru, line 5: Looks better. On thing 

Maybe i am wrong, but eavesdropping protection is usually property of privacy.

Confidentially is usually ethical duty of information recipient to don’t re-share information he got from sender. 

For example `doctor` should keep `patient` information confidential and keep place of medical examination private.

Some protocols can support confidentially by nature. For example, CL-signatures.

In DID Comm i think we can only achieve absence of additional threats to confidentiality like unexpected non-repudiation.

TelegramSam:
Discussed WG 20220207. Merged given changes made to feedback.

PR #328, opened 2022-02-07
fix #265 - remove mention of JSON-LD

### LINKED ISSUE #265: Remove mention of support for JSON-LD due to lack of implementations
I think we remove mention of JSON-LD in the spec and this issue is here to discuss that. I've not seen people using it in any way and given protocols require logical processing of messages above and beyond the semantics of the data model, I don't see what additional value this is buying us. In this case, JSON-LD seems to serve as more of a disadvantage than an advantage here.

dhh1128:


TelegramSam:
discussed WG 20220207

PR #329, opened 2022-02-07
Address a few comments from AnomalRoil

dhh1128:


mccown:
Those look good.  Thanks for submitting.  This one's straightforward, so we should be able to approve it in the meeting today.

TelegramSam:
Discussed WG 20220207.

PR #340, opened 2022-03-07
Corrects Indent and declares role of protocols

### LINKED ISSUE #138: Need explanation of DIDComm's role as a foundation for higher-level protocols
Some who approach DIDComm may view it simply as a transport mechanism, with the purpose being to enable arbitrary messages to flow securely. That is not wrong, exactly -- but it's incomplete. A major purpose of many features in DIDComm (e.g., timing headers, threading, etc) is to facilitate higher-level protocols. If you read the DIDComm spec and don't get that, you've missed something really important.

Our current spec doesn't surface this truth, and doesn't provide guidance about how to write a protocol atop DIDComm.

I suggest we add a couple paragraphs to the core spec explaining the general theory, and then add some content to the Guide that explains how to take advantage of DIDComm features to build good protocols.

TelegramSam:
Fixes #138

### DIFF
--- a/docs/spec-files/protocols.md
+++ b/docs/spec-files/protocols.md
@@ -1,6 +1,6 @@
-### Protocols
+## Protocols
 
-DIDComm serves as a foundational layer for the development of protocols. This spec does not include protocols themselves, but describes the basic requirements and foundational elements.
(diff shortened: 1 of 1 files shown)


TelegramSam:
Discussed WG 20220307. Will add reference to spec defind protocols and the creation of new protocols outside the spec.

Issue #252, opened 2021-08-13
Correlation issue in case of multiplex encryption for multiple-recipients 


ashcherbakov:
The issue is closely related to #218 and #219, and is actually being discussed in the scope of #218 (see [comment](https://github.com/decentralized-identity/didcomm-messaging/issues/218#issuecomment-889145630)).
The issue seems quite fundamental and important so that it's worth creating a dedicated ticket and be reflected in the Spec.

It looks like there is a potential privacy disclosure and correlation issue if we use multiplex encryption for multiple DIDs (there is no issue to encrypt to multiple keys for one DID). A correlation is possible even if the sender ID is protected.
Example of multiplex encryption: [ecdh-1pu-04#appendix-B](https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#appendix-B)
Please find the issue description below, and the options how we can deal with this.

## Issue
DIDComm supports protection of the sender ID (see #219  for details; it can be achieved by encrypting the `skid` or by calling anon+auth crypt). But if we want to send a message to multiple recipients (N-wise DID for example) and use multiplex encryption (as in [ecdh-1pu-04#appendix-B](https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#appendix-B)) for multiple DIDs, then all target DIDs are open and not protected.
So, a mediator node (the one in front of final recipients) will know that all N DIDs have a connection after a number of exchanged messages.

**Example:**
- Alice wants to send a message to Bob and Carol via multiplex encryption. Alice builds and encrypts the message, and sends it to Bob's mediator and Carol's mediator.
- Bob's mediator sees Bob's DID and Carol's DID in the recipient list, so it concludes that **there is a connection between that DIDs.** The same for Carol's mediator.
- Then Bob wants to send a message to Alice and Carol via multiplex encryption. Bob builds and encrypts the message, and sends it to Alices's mediator and Carol's mediator. 
- Carol mediator sees Alices's DID and Carol's DID in the recipient list, and this is the same Carol's DID as in the previous message. So, Carol mediator knows that **all three DIDs have a connection**.
- After a number of exchanged messages all mediators (the ones in front of final recipients) know that all three DIDs have a connection.

## Options how to deal with the issue

**Option 1: additional anoncrypt wrapping the message for the final recipient in case of multiplex-encryption for multiple DIDs**
[...]

PR #341, opened 2022-03-07
Reduce to single recipient

TelegramSam:
Adds text needed by #252, after which it will be marked as deferred.

### DIFF
--- a/docs/spec-files/message_structure.md
+++ b/docs/spec-files/message_structure.md
@@ -40,7 +40,7 @@ When persisted as a file or attached as a payload in other contexts, the file ex
 
 ### DIDComm Encrypted Messages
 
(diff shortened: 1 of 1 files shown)


TelegramSam:
Needs Discussion - How do we handle in the text the difference between a logical recipient and each key used for the logical recipient?

TelegramSam:
Discussed WG 20220307. Use Recipient DID and Recipient key(s) instead of just recipient to clarify the language.

TelegramSam:
WG continued: encrypted to a single party (but with potentially multiple keys) (and delivery route) at a time. 

The to header (in plaintext) is used as a courtesy to help recipients to know who else might have received the same message.

TelegramSam:
killed in favor of #360

PR #342, opened 2022-03-07
Clarifies DID Rotation

### LINKED ISSUE #291: DID Rotation section needs more details
Here are a few comments about the [DID rotation section](https://github.com/decentralized-identity/didcomm-messaging/blob/9039564e143380a0085a788b6dfd20e63873b9ca/docs/spec-files/message_structure.md#did-rotation):

> the new DID is put into immediate use encrypting the message

This could be a MUST / SHALL requirement? 
Or is it more of a SHOULD?

It latter says "The new DID and associated DID Document information SHOULD be used for further communication", so SHOULD is maybe best.

Also, does that mean we are mandated to encrypt the message when doing DID Rotation? 
What if we are only using Signed DIDComm Messages?

> from_prior: REQUIRED. A JWT, with with sub: new DID and iss: prior DID, with a signature from a key authorized by prior DID.

with with sub --> typo

This part needs extra work: it's not as well formatted as the rest and it is no clear what needs to be signed exactly. 
Is it just the `iss` field? Is it the `from_prior` field? 
But then where is the signature contained?

It is in general not clear where the signature is stored in that case. It seems this isn't necessarily a DIDComm Signed Message as per the spec, or is it? Is that "yet another type" of message not listed in the message type section?

> If multiple messages are received to containing the rotation headers after being processed by the recipient, they MAY be ignored.

Typo "messages are received to containing".

Also I'm guessing they "MAY be ignored" **only is they are the same**, no?

Also keeping track of such things means it would encourage a stateful usage of DIDComm, which seems unnecessary.
I would recommend rephrasing into something more like:
"If messages are received containing rotation headers with `sub` entries that correspond to the currently known DID for that sender, they MAY be ignored."

### § 3.2.1 JWT Details

It isn't super clear from the text that this is referencing the JWT used in the DID Rotation.
Maybe name it explicitly? "The `from_prior` JWT is constructed..."

It is not specified what the `iat` is supposed to be in the text. Instead it just says in the example:

> "iat": 1516239022 //datetime of the rotation, not message

This should be made explicit.
Also are we to verify that date in some way? 

Are we accepting times in the future? 
Shall we be waiting until then to rotate if so? (Which would contradict the "put into immediate use" from above.)
Or shall we reject DID rotation specifying a `iat` date in the future?

### § 3.2.3 Rotations Limitations

> This rotation method only supports the case where a new DID is used, replacing an old DID which is no longer used. Adjustments to DIDs used between different parties that does not fit this narrow use are expected to define a separate protocol to do so.

So, what happens in practice when someone "updates" their current DID and we are facing a DID we "already know" but using a key, e.g., that isn't in the version of that DID we're knowing about?

DID docs are versioned, so we should prolly have some verbiage in the DIDComm spec about how these versions are compared / used in DIDComm to ensure we're up to date wrt. each other's DID Docs...

TelegramSam:
Fixes #291

### DIFF
--- a/docs/spec-files/message_structure.md
+++ b/docs/spec-files/message_structure.md
@@ -136,21 +136,21 @@ Headers can be simple (mapping a header name to an integer or a string) or struc
 
 ### DID Rotation
 
-DIDComm Messaging is based on [DIDs](https://www.w3.org/TR/did-core/) and their associated [DID Documents](https://www.w3.org/TR/did-core/#dfn-did-documents). Changes to keys and endpoints are the concern of each [DID](https://www.w3.org/TR/did-core/) method and are utilized but not managed by DIDComm Messaging. [DID Rotation](#did-rotation) serves a very specific and narrow need to switch from one [DID](https://www.w3.org/TR/did-core/) method to another. This is very common at the beginning of a new DIDComm Messaging relationship when a public [DID](https://www.w3.org/TR/did-core/) or a temporary [DID](https://www.w3.org/TR/did-core/) passed unencrypted is rotated out for a [DID](https://www.w3.org/TR/did-core/) chosen for the relationship. As rotation between one [DID](https://www.w3.org/TR/did-core/) and another is outside the scope of any [DID](https://www.w3.org/TR/did-core/) method, the details of [DID Rotation](#did-rotation) are handled within DIDComm Messaging itself.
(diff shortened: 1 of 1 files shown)


TelegramSam:
Discussed WG 20220307. Need advice on when to perform a rotation. Avoid multi-threaded situations where multiple prototcols are in action at the same time.

Care should be taken to avoid ambiguity. 

Avoid casual key rotation. seek a quiet period.

consider MUST for discarding messages from old DID, include cybersecurity reasons in explanation.

dhh1128 (approved):
LGTM

PR #361, opened 2022-04-09
remove meaningless typ

### LINKED ISSUE #317: `typ` specified in plaintext message format?
Several places in the spec have `typ` specified in the plaintext form of the message. I think this is quite redundant, as you already can understand the format because you have to look there to see the `typ` attribute.

TelegramSam:
Fixes #317


mccown:
Approved in the DIDComm WG on 20220411.

PR #363, opened 2022-04-11
Change from dash to underscore for attributes

TelegramSam:


mccown:
These are good updates. From what I'm reading, all languages accept underscores while some have issues with hyphens, so using underscores is a more universal solution.

mccown:
Approved in the DIDComm WG on 20210418

PR #365, opened 2022-04-14
fix https as transport over http

TelegramSam:
This acts as a companion to @mccown's adjustment to HTTPS for all examples. It fixes the requirement for HTTPS for transports. This addresses concerns about mobile agents and the exceptions mobile platforms are required to make if a regular HTTP connection is used. It also addresses the perception of HTTPS always being a best practice, even if in this case it results in double encryption.


mccown:
Approved in the DIDComm WG on 20210418

PR #367, opened 2022-04-15
improve ACKs

dhh1128:


TelegramSam:
Discussed in WG 20220418. Nothing defined for what happens if you don't have any record of that exchange (maybe you didn't retain it after processing), but as ack responses are optional, maybe this is ok.

mccown:
Approved in the DIDComm WG on 20220418

PR #370, opened 2022-04-18
Specify anoncrypt OR authcrypt

### LINKED ISSUE #208: Add language to forbid mixing anon/authcrypt in ONE JWE
We should add normative language in the spec that says don’t mix anoncrypt with authcrypt for multiple recipients in one JWE. NOTE: it should still be ok to have nested JWEs that use different key wrapping algorithm, e.g., authcrypt and anoncrypt.

TelegramSam:
Fixes #208


dhh1128 (approved):

mccown:
I like this clarification.  Given the consensus, I'll go ahead and merge.

PR #375, opened 2022-04-22
fix #269

### LINKED ISSUE #269: Refactor "Key IDs kid and skid headers references in the DID document" section
- we should not define apu, apv in DIDComm Messaging, in 1PU-V4 it is just RECOMMENDED, not required.
- we should have only one place to find the sender key which must be `skid`, with the format that is already defined our spec
- improve readability and removing redundant normative statements see below.

imo, we don't need:
> Keys used by DIDComm envelopes MUST be sourced from the DIDs exchanged between two agents.

imo, we cannot make examples normative:
> The corresponding kid header for this recipient MUST have did:example:jklmnopqrstuvwxyz1#key-1 as value.

here another one:
> the unpacking process on his end MUST resolve the skid protected header value using Alice’s DID doc’s KeyAgreement[0]

and here:
> KeyAgreement[0] is a reference id, it MUST be resolved from the main VerificationMethod[] of Alice’s DID document (not shown in the example).

dhh1128:

### DIFF
--- a/docs/spec-files/encryption.md
+++ b/docs/spec-files/encryption.md
@@ -62,7 +62,7 @@ DIDComm Messaging's guarantees with respect to man-in-the-middle attacks are eas
 
 #### Key IDs `kid` and `skid` headers references in the DID document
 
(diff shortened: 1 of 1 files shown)


TelegramSam:
This fixes the overuse of normative language pointed out in #269. I vote merge.

PR #384, opened 2022-04-29
fix #381

### LINKED ISSUE #381: Attachment examples @id to id
The examples that include attachments have an `@id` for their ids whereas the attachment id is actually an `id` field. Examples need to be updated to proper property

dhh1128:


dhh1128:
Fixes #381

mccown:
Approved in the DIDComm WG on 20220502.

Issue #236, opened 2021-07-21
need clarification about encrypted(encrypted(payload)) in mime types table


dhh1128:
We don't describe the `encrypted, then encrypted` scenario that's described in the mime types table. We need to explain ourselves more.

I believe our intention with it is always `anoncrypt(authcrypt(data))` option, where “data” might be a signed message or a plaintext message. It is an unnatural behavior that is only used in one case, which is to hide the key of the sender from the final mediator in a delivery route.

Do others agree that this is what we meant? If so, I will raise a PR to add some explanatory language about it.

dhh1128:
tagging @vimmerru

mirceanis:
That's the way I interpreted it too.
In our implementation we define the possible message "packing" options as:
* plain
* signed
* anoncrypt
* authcrypt
* anoncrypt + signed
* anoncrypt + authcrypt

vimmerru:
@dhh1128 One additional point, `anoncrypt(authcrypt(data))` and `authcrypt(data)` and `anoncrypt(authcrypt(sign(data)))` have one mime type. If we force only few and exact wrappings should it be different mime types?

Also what we should do if authcrypt, sign and plain messages here use different `from`? Or if `from` missed in plain text? How `trust` is combined when we remove authcrypt, anoncrypt or sign envelopes? I think spec should describe it also.

dhh1128:
>have one mime type. If we force only few and exact wrappings should it be different mime types?

I don't think so. We don't want people to know what's inside an outer layer of encryption; from the perspective of someone who may or may not be able to decrypt it, all of these should look the same. It is only someone who is capable of peeling away the wrappers that should know more detail.

ashcherbakov:
We should probably address the following the Spec: [Possible-nested-message-combinations](https://hackmd.io/aD4TzKRhT1ma2ERl6EKU_g?view#Possible-nested-message-combinations)

awoie:
From my perspective, all DIDComm message types are just tools for higher-level protocols. IMO, it would make sense to recommend packing methods to achieve certain properties (e.g., authcrypt(data) as default for identifier auth, anoncrypt(authcrypt(data)) to hide sender  etc.) but let the higher-level protocols to decide which properties are needed, e.g., sender authentication.

To your question @dhh1128, yes, it was the intend to use anoncrypt(authcrypt(data)) for sender anonymity. And I also agree, we could improve language and in particular the table a little bit better.

Issue #219, opened 2021-07-05
Need more details about protected `skid`


dhh1128:
A developer wanting to implement the spec read this section:

>https://identity.foundation/didcomm-messaging/spec/#protecting-the-skid-header

His comment was that there is not enough detail to implement this mechanism. I agree with him. I think we need an example of what is meant by this sentence:

>MAY include an encrypted_skid header with a value based on the encryption of skid using ECDH-ES Z computation of the epk and the recipient’s key as the encryption key.

dhh1128:
tagging @vimmerru

awoie:
We agreed to revise this section in a separate PR: https://github.com/decentralized-identity/didcomm-messaging/pull/185#issuecomment-860928028 . AFAIK, the decision was to not encrypt the `skid` and using `encrypted_skid`. We would rather add non-normative language how protection can be achieved by other means, e.g., pairwise-sender-DIDs, or anoncrypt(authcrypt(msg)).

ashcherbakov:
@kdenhartog it looks like this issue should have `PR Needed` label, as the current spec doesn't provide full or up-to-date information related to `skid`. Do you agree?

kdenhartog:
Looks like I didn't think about this enough when triaging. Thanks for the suggestion updated that.

vimmerru:
We discussed it internally with @ashcherbakov and got some recommendations you can consider:

1. Proposed way to hide sender (use `anoncrypt(authcrypt())`) doesn't look optimal. It will require `n` additional ECDH operations and key wrappings, additional content encryption and size of the message will  growth significantly. It obviously complicates protocol (forces message layering) and opens possible misusage ways.
1. JWE/ECDH-1PU specifications doesn't specify how content of skid header should look like. It should only allow recipient to determine the key of sender. So i think we don't need dedicated `protected_skid` header and can just use `skid` header to complain JWE/ECDH-1PU.
1. For pairwise or even n-wise communication when recipient party knows in advance the sender DID (or possible sender DIDs) we can use `H(KeyURI, IVkw, Tagkw)` as SKID. Recipient can iterate over sender DIDDoc/DIDDocs and find a key that corresponds to hash. 
1. For open communication KID can be anoncrypted JWE that uses same recipient keys. Potentially we can consider re-usage of ephemeral keys or other secrets used for ECDH-1PU to reduce amount of crypto operations, but it require more time for analysis.


[...]

PR #386, opened 2022-05-09
fix 236

dhh1128:


dhh1128:
Fixes #219
Fixes #236

TelegramSam:
Discussed WG 20220509. Soften MUST NOT to SHOULD NOT for `authcrypt(sign(plaintext))` with a warning about `anoncrypt(authcrypt(sign(plaintext)))` not being valid, and must use `anoncrypt(sign(plaintext))` being the thing to use in that case.
No objections to merging this with proposed changes.

PR #387, opened 2022-05-09
fix #217

### LINKED ISSUE #217: Need details about invalid curve and weak point attack resistance
The section about key negotiation specifies an algorithm, but not best practices for using it. Without additional guidance, DIDComm could be vulnerable to the same sort of theoretically-correct-but-practically-flawed implementations that caused this well known article to be written:

https://www.infoworld.com/article/3184582/critical-flaw-alert-stop-using-json-encryption.html

dhh1128:

fixes #217


mccown:
Discussed and approved in the DIDComm WG on 20220409.

PR #391, opened 2022-05-11
consistency  checks

### LINKED ISSUE #237: Should implementation check consistency of 'from', 'to', 'skid' and similar headers for authcrypt, anoncrypt, sign envelopes and plaintext?
I suggest imagine multiple situations to think and answer:

1. We have Signed message from Alice to Bob, but `from` field in plaintext is empty or different. 

- Should implementation check it and return error for this case?
- If not should implementation not only unwrap sign envelope and return plaintext, but return additional metadata with some information about trust received from sign envelope?

2. We have message in `anoncrypt(authcrypt(plaintext))` envelopes, but set of `to` dids/keys in anoncrypt, authcrypt and maybe plaintext is different.

- Should implementation check it and return error for this case?

3. We have `authcrypt(plaintext)` message encrypted for multiple keys. Bob owns multiple of this case on this device.

- Should implementation check auth for all keys owned by Bob?
- What it should do if one key is correctly wrapped, but second not?
- Should implementation return some metadata about what keys where checked?

4. We have `anoncrypt(authcrypt(plaintext))` message to Bob. anoncrypt and authcrypt parts have different `to`, but Bob still can decrypt it as he owns keys for both parts.

- Should implementation check it and return error for this case?
- If not should it return additional metadata about keys used for decrypt?

It's just examples of such cases i can imagine significantly more similar corner cases. Ideally specification should define process of unwrapping every envelope and checks it should perform after it.

TelegramSam:
Fixes #237

### DIFF
--- a/docs/spec-files/message_structure.md
+++ b/docs/spec-files/message_structure.md
@@ -155,9 +155,11 @@ Headers can be simple (mapping a header name to an integer or a string) or struc
 
 When messages are combined into layers as shown above in the Media Types table, various attributes must be checked for consistency by the message recipient.
 
-- The `from` attribute in the plaintext message must match the `skid` attribute in the encryption layer.
-- The `to` attribute in the plaintext message must contain the `kid` attribute of an encrypted message.
-- The `from` attribute in the plaintext message must match the signer's `kid` in a signed message.
+- The `from` attribute in the plaintext message MUST match the `skid` attribute in the encryption layer.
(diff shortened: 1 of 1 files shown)


dhh1128 (changes requested):
I agree with the content but suggest two minor changes:

1. Make the word "must" appear in all caps.
2. Add a sentence: "When one of these checks fails, the result MUST be an error so clients know that the trust choices in the message packaging are inconsistent."

mccown:
Approved in the DIDComm WG on 20220516.

PR #393, opened 2022-05-12
ecdh-es header section

### LINKED ISSUE #256: Need more info about headers in anoncrypt (ECDH-ES) case
The current spec provides details about common protected headers for authcrypt (ECDH-1PU) case: [link](https://identity.foundation/didcomm-messaging/spec/#ecdh-1pu-key-wrapping-and-common-protected-headers).
But there is no such info for anoncrypt (ECDH-ES) case. The general JWE/JWA specs don't specify what must be set for some of the headers, as it's up to application (see [link](https://datatracker.ietf.org/doc/html/rfc7518#section-4.6.2)).

So, we need to specify the following for anoncrypt (ECDH-ES) in the spec:
1) What common protected headers must be used  (`apv`, `apu`, `alg`, `epk`?)
2) What should be used as `apv` value
3) What should be used as `apu` value
4) Whether `alg` and `epk` must be in protected or per-recipient unprotected headers 
   - specific case: multiplex-encryption for multiple keys of potentially different types, where we may generate multiple epks for every recipient key type.

TelegramSam:
Fixes #256

I'm not sure this is complete, in spite of filling the requirements in #256 Review is needed here for correctness.

### DIFF
--- a/docs/spec-files/encryption.md
+++ b/docs/spec-files/encryption.md
@@ -146,6 +146,21 @@ Even though `apu`/`apv` are not mandatory JWE recipients headers, they are requi
 
 A final note about `skid` header: since the 1PU draft [does not require](https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#section-2.2.1) this header, authcrypt implementations MUST be able to resolve the sender kid from the `apu` header if `skid` is not set.
 
+#### ECDH-ES key wrapping and common protected headers
+
+When using anoncrypt, any of the valid content encryption algorithms may be used. To meet this requirement, JWE messages MUST use common `epk`, `apv` and `alg` headers for all recipient keys. They MUST be set in the `protected` headers JWE section.
+
+As per this requirement, the JWE building must first encrypt the payload, then use the resulting `tag` as part of the key derivation process when wrapping the `cek`.
+
+To meet this requirement, the above headers are defined as follows:
+
+* `epk`: generated once for all recipient keys. It MUST be of the same type and curve as all recipient keys since kdf with the sender key must be on the same curve.
+  - Example: `"epk": {"kty": "EC","crv": "P-256","x": "BVDo69QfyXAdl6fbK6-QBYIsxv0CsNMtuDDVpMKgDYs","y": "G6bdoO2xblPHrKsAhef1dumrc0sChwyg7yTtTcfygHA"}`
+* `apv`: this represents the recipients' `kid` list. The list must be alphanumerically sorted, `kid` values will then be concatenated with a `.` and the final result MUST be base64 URL (no padding) encoding of the SHA256 hash of concatenated list. 
+* `alg`: this is the key wrapping algorithm, ie: `ECDH-ES+A256KW`.
+
+Note: `apu`will not be present when using ECDH-ES.
+
 #### Examples
 
 While the details of encrypting a JWM into a JWE are included in the [JWM spec](https://tools.ietf.org/html/draft-looker-jwm-01), a few examples are included here for clarity. See section [Appendix C.3](#c3-didcomm-encrypted-messages).
--- a/docs/spec-files/encryption.md
+++ b/docs/spec-files/encryption.md
@@ -6,7 +6,7 @@ DIDComm Messages are encrypted with the keys of a single DID. A message being se
 
 DIDComm supports two types of message encryption: Authenticated Sender Encryption ("authcrypt") and Anonymous Sender Encryption ("anoncrypt"). Both forms are encrypted to the recipient DID. Only authcrypt provides direct assurances of who the sender is. Each encrypted message MUST use either authcrypt or anoncrypt.
 
-The encrypted form of a JWM is a JWE. The JOSE family defines [JSON Web Algorithms](https://tools.ietf.org/html/rfc7518) (JWAs) which standardize certain cryptographic operations that are related to preparing JOSE structures. For the purposes of interoperability, DIDComm messaging does not support all JWAs; rather, it takes a subset of the supported algorithms that are applicable for the following cases around secure messaging. These supported algorithms are listed in tables later in the spec.
+The encrypted form of a JWM is a JWE in General JSON Format. The JOSE family defines [JSON Web Algorithms](https://tools.ietf.org/html/rfc7518) (JWAs) which standardize certain cryptographic operations that are related to preparing JOSE structures. For the purposes of interoperability, DIDComm messaging does not support all JWAs; rather, it takes a subset of the supported algorithms that are applicable for the following cases around secure messaging. These supported algorithms are listed in tables later in the spec.
 
 
 #### Sender Authenticated Encryption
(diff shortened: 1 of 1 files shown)


dhh1128 (approved):
This looks good to me, except for one minor comment about formatting.

However, I would like some people with greater confidence on the cryptography to confirm what we are saying here. Tagging @vimmerru, @ashcherbakov, and @Baha-sk

docs/spec-files/encryption.md
```diff
@@ -146,6 +146,21 @@ Even though `apu`/`apv` are not mandatory JWE recipients headers, they are requi
 
 A final note about `skid` header: since the 1PU draft [does not require](https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#section-2.2.1) this header, authcrypt implementations MUST be able to resolve the sender kid from the `apu` header if `skid` is not set.
 
+#### ECDH-ES key wrapping and common protected headers
+
+When using anoncrypt, any of the valid content encryption algorithms may be used. To meet this requirement, JWE messages MUST use common `epk`, `apv` and `alg` headers for all recipient keys. They MUST be set in the `protected` headers JWE section.
```
dhh1128, line 151: I'm not sure why "protected" is formatted as code here. Is there actually a section of the output data structure that is tagged in this way? In other places, I've just seen the phrase "protected headers" without any special formatting...

Baha-sk: > To meet this requirement, JWE messages MUST use common `epk`, `apv` and `alg` headers for all recipient keys. They MUST be set in the `protected` headers JWE section.

This is only true for a single recipient envelopes, which can be set as [compact serialized JWE format](https://datatracker.ietf.org/doc/html/rfc7516#section-7.1) (which supports a single recipient only) or [Flattened JSON serialization](https://datatracker.ietf.org/doc/html/rfc7516#section-7.2.2) format (it also supports a single recipient only). In these formats, the recipient specific headers can be merged into the Protected headers and `recipients` section is skipped.

For multi-recipient envelopes, each recipient must have their own `header` and `encrypted_key` section as per the [JWE serialization formatting rules](https://datatracker.ietf.org/doc/html/rfc7516#:~:text=single%2Drecipient%20case.-,7.2.1,-.%20%20General%20JWE%20JSON). 

I'd probably avoid sharing the same encrypted key (encrypted cek) value with all recipients in the case where recipients represent different entities (ie if the recipients list comprises my different devices keys only, then it's ok to share, but if it represents different users, then it's not ok to share for additional protection). In the end, each recipient must own the private key used in the ECDH key wrapping in order to unwrap the cek. This is why each recipient must have their own `header` section under `recipients` in the general JSON format. 

In the case of a single recipient, then compact or JSON Flattened format can be used. This is the only case I see where `epk`, `apv` and `apu` can be merged into the `protected` headers.

dhh1128: @Baha-sk : Thanks for the comments.

We've decided that we will use multi-recipient features to send the same message to multiple devices belonging to a single party (e.g., Bob has 3 agents); however, we will not use multi-recipient features to send the same message to Bob, Carol, and David. The latter doesn't make sense, since Carol and David may have different key types from Bob, and will certainly have different mediators and routes.

Does this alter your concerns in any way?

Baha-sk: > @Baha-sk : Thanks for the comments.
> 
> We've decided that we will use multi-recipient features to send the same message to multiple devices belonging to a single party (e.g., Bob has 3 agents); however, we will not use multi-recipient features to send the same message to Bob, Carol, and David. The latter doesn't make sense, since Carol and David may have different key types from Bob, and will certainly have different mediators and routes.
> 
> Does this alter your concerns in any way?

Thanks for the clarification @dhh1128 

My only concern is mostly about spec compliance to be honest. 

My understanding from the JWE serialization sepc I linked above is it entails 1 key per recipient regardless of who owns the key. 1 user having 3 device keys = 3 recipients and therefore must use General JSON format.


If I understand correctly, what we're aiming for here is to have a hybrid JWE model: Use a JSON Flattened format with multi recipient keys. I'm not sure it's doable since key Wrapping (and its ECDH key derivation) must be done separately for each key which will generate unique `encrypted_key` entries. How can this list be formatted in a single recipient JSON Flattened format?

Baha-sk: >We've decided that we will use multi-recipient features to send the same message to multiple devices belonging to a single party (e.g., Bob has 3 agents);

if this means sending separate JWE envelopes for each device, then we're ok. I think the misunderstanding came from sending the same envelope to all three devices at once.

I think we can clarify this by saying that `each device will obtain its unique envelope targeted to its respective key` or something along this line.

Baha-sk (approved):
it looks good to me. The minor clarification discussion above is helpful, but not necessary if we don't get into too much detail.

dhh1128:
Wait, wait. I AM saying that we send one envelope to 3 different devices that all belong to the same user. If that's not what Sam's saying then we have a very surprising misalignment.

IMO, it is NOT okay to send a different encryption envelope to each device. That defeats the whole goal of having multiplex encryption.

Baha-sk:
> Wait, wait. I AM saying that we send one envelope to 3 different devices that all belong to the same user. If that's not what Sam's saying then we have a very surprising misalignment.
> 
> IMO, it is NOT okay to send a different encryption envelope to each device. That defeats the whole goal of having multiplex encryption.

JWE has 3 formats as described earlier:
1. Compact: the 5 `.` separated base64URL encoded sections representing a JWE for a single recipient (1 key).
2. Flattened JSON: a JSON format of a JWE envelope for a single recipient (1 key) message. 
These two formats support merging `recipients[0].header` into the `protected` section since there's only 1 recipient (to answer your original question: yes `protected` is a special section representing the protect headers in a JWE envelope).

3. General JSON Format: The same format as Flattened JSON but with multiple recipients (multiple keys) each with its own headers in `recipient[x].header`. Recipient headers usually include `epk`, `apu, `apv`, etc. These cannot be merged in to the `protected` headers common section of the JWE since each key has a unique key wrapping (and therefore ECDH-ES computation) of the cek.

- if we want to support multiple devices, each with its own key, then we have to use option 3. Each device is treated as a separate recipient with its own key.
- if, on the other hand, we want to support sending the same envelope to, say, all 3 devices of a single user (recipient), then all three devices must share the same recipient key. Only then we can use option 1 or 2.

docs/spec-files/encryption.md
TelegramSam: TODO: Need to remove the tick marks around protected as it's not a literal value in the JWE.

TelegramSam: Nevermind. keep ticks, remove 'headers' from 'protected' _headers_ in both auth and anoncrypt sections.

TelegramSam:
Needed: `The encrypted form of a JWM is a JWE. ` Add JWE, in the General JSON Format.

dhh1128 (approved):
LGTM

mccown:
This was approved in the recent DIDComm WG on 20220516 pending a few modifications, which have been added.  With those changes, it's ready for merging.

PR #396, opened 2022-05-21
fix: kebab instead of snake case on ping-response

dbluhm:
Signed-off-by: Daniel Bluhm <dbluhm@pm.me>

### DIFF
--- a/docs/spec-files/trustping.md
+++ b/docs/spec-files/trustping.md
@@ -34,13 +34,13 @@ creates a `ping` message like this:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


dhh1128 (commented):
Do we know that the kebab-case version is the one that's broadly implemented?

dbluhm:
Fair question. For the 1.0 version of this protocol, the message type name used is `ping_response`. I am not aware of any 2.0 implementations yet, personally. That being said, the snake casing of this message type name has always been an odd one out and I'd strongly advocate for switching to kebab case with a 2.0 version of this protocol, even if there's an implementation or two already out there.

dbluhm:
Also worth pointing out that the example message given a few lines after the changes made in this PR show `ping-response`. Whichever way we go, we'll have to make a correction to be internally consistent.

PR #399, opened 2022-05-23
Correct OOB Attachments to Optional

### LINKED ISSUE #397: Should attachments be required on OOB Invitations?
The current text for the `attachments` attribute reads:

> `attachments` - REQUIRED for OOB usage. An array of attachments that will contain the invitation messages in order of preference that the receiver can use in responding to the message...

After some discussion with Sam, it seems the intent for OOB invitations is that instead of always having an attachment with an explicit message or set of messages, it should be possible to just specify a goal code that initiates some flow of messages. To support that use case, attachments should be marked as "OPTIONAL" rather than "REQUIRED."

cc @TelegramSam

TelegramSam:
Fixes #397 

@dhh1128 this requirement for attachments was introduced during a recent commit to formalize field names. This corrects it to Optional.


dhh1128 (approved):
Let's merge this!

mccown:
Discussed and approved in the DIDComm WG on 20220523 along with the requested review.

PR #407, opened 2022-08-08
Updates a few version numbers from 1.0 to 2.0

mccown:
Some of the examples used version 1.0 and I updated those to version 2.0.

### DIFF
--- a/docs/spec-files/empty.md
+++ b/docs/spec-files/empty.md
@@ -4,13 +4,13 @@ Sometimes, only headers need to be communicated; there is no content for the bod
 
 The [PIURI](#protocol-identifier-uri) for this protocol is:
 
-    https://didcomm.org/empty/1.0
+    https://didcomm.org/empty/2.0
 
 The `empty` message has no semantic meaning. The message's only purpose is to allow the transfer of message headers. 
 
 ```JSON
 {
-  "type": "https://didcomm.org/empty/1.0/empty",
+  "type": "https://didcomm.org/empty/2.0/empty",
   "id": "518be002-de8e-456e-b3d5-8fe472477a86",
   "from": "did:example:123456",
   "body": {}
--- a/docs/spec-files/empty.md
+++ b/docs/spec-files/empty.md
@@ -4,13 +4,13 @@ Sometimes, only headers need to be communicated; there is no content for the bod
 
 The [PIURI](#protocol-identifier-uri) for this protocol is:
 
-    https://didcomm.org/empty/2.0
+    https://didcomm.org/empty/1.0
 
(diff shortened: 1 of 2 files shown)


docs/spec-files/empty.md
```diff
@@ -4,13 +4,13 @@ Sometimes, only headers need to be communicated; there is no content for the bod
 
 The [PIURI](#protocol-identifier-uri) for this protocol is:
 
-    https://didcomm.org/empty/1.0
+    https://didcomm.org/empty/2.0
```
TelegramSam, line 7: empty is newly defined in this spec - so there's no 1.0 anywhere else.

docs/spec-files/problems.md
```diff
@@ -164,10 +164,10 @@ To troubleshoot routing issues, DIDComm offers a header, `trace`. Any party that
 
 ```json
 {
-  "type": "https://didcomm.org/trace/1.0/trace_report",
+  "type": "https://didcomm.org/trace/2.0/trace_report",
   "pthid": "98fd8d72-80f6-4419-abc2-c65ea39d0f38.1",
   "handler": "did:example:1234abcd#3",
-  "traced_type": "https://didcomm.org/route/1.0/forward",
```
TelegramSam, line 170: the protocol name is `routing` as an additional fix.

docs/spec-files/empty.md
```diff
@@ -4,13 +4,13 @@ Sometimes, only headers need to be communicated; there is no content for the bod
...
 
 The `empty` message has no semantic meaning. The message's only purpose is to allow the transfer of message headers. 
 
 ```JSON
 {
-  "type": "https://didcomm.org/empty/1.0/empty",
+  "type": "https://didcomm.org/empty/2.0/empty",
```
TelegramSam, line 13: same as comment above. This didn't need a version rev.

docs/spec-files/problems.md
```diff
@@ -164,10 +164,10 @@ To troubleshoot routing issues, DIDComm offers a header, `trace`. Any party that
 
 ```json
 {
-  "type": "https://didcomm.org/trace/1.0/trace_report",
+  "type": "https://didcomm.org/trace/2.0/trace_report",
```
TelegramSam, line 167: this is fuzzy as it's not a defined protocol name. I think leaving 2.0 here is the right answer.

docs/spec-files/empty.md
mccown: That makes sense.  When I saw the "1.0", I had associated it with the spec's version number instead of the protocol's version number.  I'm happy with whichever is most clear to everyone.

docs/spec-files/empty.md
mccown: My comment is also the same as above.  :-)

PR #411, opened 2022-10-05
proposal for filesystem transport extension

dhh1128:


TelegramSam:
Your bandwidth comments immediately made me think of DIDComm over AWS Snowball as a transport. (I'm pretty sure this scheme would work with that.)

Also, Fedex Bandwidth XKCD link: https://what-if.xkcd.com/31/

extensions/filesystem_transport/main.md
```diff
@@ -0,0 +1,77 @@
...
+
+Aside from the extension, the remainder of the filename of an individual message is arbitrary. However, it MAY include the value from the message's `id` header to guarantee uniqueness.
+
+#### Ordering
+
+When more than one DIDComm message is found in a single folder, receivers MUST process them in a predictable order to simulate the order-of-arrival semantics that are taken for granted in most other transports.
+
+Two ordering semantics are supported. The simplest and default is to process files in order of ascending last modification date. Usually this suffices. However, there may be corner cases where this is inadequate. Therefore, it is also possible for senders to force an arbitrary order by naming files with an __ordinal prefix__ that consists of digits followed by a double hyphen, as in `01--msg.dcem`, `02--msg.dcem`. Leading zeros are optional; the double hyphen is required to distinguish this convention from names based on UUIDs and other identifiers that use digits and hyphens. When filenames of messages begin with an ordinal prefix, messages MUST be processed by a receiver in the numeric (NOT lexicographic) order implied by the prefixes.
```
rodolfomiranda, line 35: For the second option of ordering, an implementor would like to have a way to avoid the sender the need to scan/read all file names in the folder to check which was the last digit used (assuming the sender does not keep track of them; I'm thinking on a mediator for example). One option is to use the unixtime mills as the ordinal prefix.

dhh1128: Excellent suggestion. I'll add a sentence about that.

dhh1128:
@TelegramSam : added a note about AWS Snowball, as I think this expands the thinking on the topic in a great way.

TelegramSam (approved):
Discussed WG 20230306

PR #426, opened 2023-01-11
'controller' in Verification-Method MUST be a DID Syntax

FabioPinheiro:
In the Verification-Method the properties controller is a `DID Syntax` https://w3c.github.io/did-core/#verification-method-properties `DID Syntax` can not have fragments. Am I seeing this correctly?

But on the example  https://identity.foundation/didcomm-messaging/spec/#b2-recipient-diddocs I see `"controller":"did:example:bob#key-x25519-1"`

### DIFF
--- a/docs/spec-files/appendix.md
+++ b/docs/spec-files/appendix.md
@@ -157,7 +157,7 @@ This section defines the sender DIDDoc.
       {
          "id":"did:example:alice#key-1",
          "type":"JsonWebKey2020",
-         "controller":"did:example:alice#key-1",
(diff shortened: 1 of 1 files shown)


FabioPinheiro:
I found this when I tried to make my code base a bit more type-safe.
I assume the examples is the thing here that is wrong

peacekeeper:
I agree the example in https://identity.foundation/didcomm-messaging/spec/#b2-recipient-diddocs is probably wrong. The "controller" should be just the DID without fragment.

TelegramSam (approved):

PR #436, opened 2023-04-03
Added support for a single serviceEndpoint 

mccown:
Added support for a single serviceEndpoint in addition to the current array of serviceEndpoint objects.  This is intended to assist with ION compatibility.


docs/spec-files/routing.md
```diff
@@ -134,13 +150,14 @@ The relevant entry in the DID document matches this format:
...
 
 `type` - REQUIRED. MUST be `DIDCommMessaging`. 
 
-`serviceEndpoint` - REQUIRED. MUST contain an ordered list of objects. Each represents a DIDComm Service Endpoint URI and its associated details. The order of the endpoints SHOULD indicate the DID Document owner's preference in receiving messages. Any endpoint MAY be selected by the sender, typically by protocol availability or preference. A message should be delivered to only one of the endpoints specified.
+`serviceEndpoint` - REQUIRED. As described above, a serviceEndpoint MUST contain a string, a map, or a set composed of one or more strings and/or maps. Each represents a DIDComm Service Endpoint URI and its associated details. The order of the endpoints SHOULD indicate the DID Document owner's preference in receiving messages. Any endpoint MAY be selected by the sender, typically by protocol availability or preference. A message should be delivered to only one of the endpoints specified.
 
-Each object has the following properties:
+If a map is used, each object has the following properties:
```
TelegramSam, line 160: a map is used in every case but string.

PR #437, opened 2023-04-04
The body attribute is now OPTIONAL

FabioPinheiro:
Make the body attribute OPTIONAL

Also tried to check for messages where the body could be removed but it seems that is already missing from there https://identity.foundation/didcomm-messaging/spec/#ping-response


FabioPinheiro:
cc @TelegramSam 

Also tried to check for messages where the body could be removed but it seems that is already missing from there https://identity.foundation/didcomm-messaging/spec/#ping-response

TelegramSam (approved):

Issue #461, opened 2025-01-23
Section 12.1.2 has a typo in the "kid" field


stormer78:
For Bob's Secrets, for each key the "kid" field name contains a trailing whitespace character which causes parsing issues.

"kid " should be "kid"

Example:
```
{
    "kid ":"did:example:bob#key-x25519-1",
    "kty":"OKP",
    "d":"b9NnuOCB0hm7YGNvaE9DMhwH_wjZA1-gWD6dA0JWdL0",
    "crv":"X25519",
    "x":"GDTrI66K0pFfO54tlCSvfjjNapIs44dzpneBgyx0S3E"
  },
```

Should become:
```
{
    "kid":"did:example:bob#key-x25519-1",
    "kty":"OKP",
    "d":"b9NnuOCB0hm7YGNvaE9DMhwH_wjZA1-gWD6dA0JWdL0",
    "crv":"X25519",
    "x":"GDTrI66K0pFfO54tlCSvfjjNapIs44dzpneBgyx0S3E"
  },
```

PR #462, opened 2025-02-06
Removes typo space in kid JSON item in examples -- per issue 461

swcurran:
Fixes the repeated typo `"kid "` in the JSON examples in the appendix, as pointed out in #461.

Note that I fixed the active spec, plus the two snapshots (2.0, 2.1), as it is clearly a mistake in all 3 sets of examples.  A picky reviewer might want to remove the "snapshot" fixes (and happy to do that), but I think it is kinda silly to keep such an obvious typo in place.

Signed-off-by: Stephen Curran <swcurran@gmail.com>

### DIFF
--- a/docs/spec-files-v2.0-snapshot/appendix.md
+++ b/docs/spec-files-v2.0-snapshot/appendix.md
@@ -67,28 +67,28 @@ This section defines the recipient private keys in JWK format.
 ```json
 [
   {
-    "kid ":"did:example:bob#key-x25519-1",
+    :"did:example:bob#key-x25519-1",
     "kty":"OKP",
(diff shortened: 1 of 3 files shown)


FabioPinheiro (changes requested):
I believe you replace `"kid "` with a empty string

swcurran:
Well, that’s embarrassing!  Fixed.  I think..

Sorry to waste your time @FabioPinheiro , let alone mine...

FabioPinheiro:
You didn't waste my time I was looking for another thing about relative-did-urls.
I indeed use those examples on my test vector, and already have it fix on my library.

Would be nice with someone merge this PR.

FabioPinheiro (approved):

